PR #10, opened 2019-05-19
Simplify local testing

### LINKED ISSUE #9: Simplify local testing
When developing, I don't necessarily want to re-run docker compose all the time to re-run tests. Having a test running setup that works for both docker-compose (in CI) and via local CLI would make that easier and allow me to easily just run the tests I want to run via `pytest`.

MrTrustworthy:
Closes #9


tests/conftest.py
```diff
@@ -17,15 +17,7 @@
...
-TEST_CONFIG = """
-[Context]
-current = local
-
-[Context.local]
-bootstrap_hosts = localhost
-bootstrap_port = 9092
-security_protocol = PLAINTEXT
-"""
+SAMPLE_CONFIG_PATH: Path = Path(__file__).parent.parent / "config" / "sample_config.cfg"
```
hfjn: You could use `sample_config_path()` from `esque.config` for this.

README.md
```diff
@@ -69,8 +69,12 @@ export PYTHONPATH=$(pwd)
...
 
 ### Run tests
+
+To run all tests, just run
+
 ```
 docker-compose up
 ```
 
+While the `docker-compose` stack is up, you can also run the tests from the CLI via `pytest tests/ --integration --local`
```
hfjn, line 126: That's actually pretty annoying because the esque docker will still be built. Should we provide a possibility to actually run the stack without esque?

hfjn (changes requested):
Two minor change requests. Looks good to me otherwise.

README.md
MrTrustworthy: Thought about that too, but i'm not sure how i'd compose the compose files (hehe) to make this work with both development options without making them more complicated

hfjn: We could create a base `docker-compose.yml` with the stack. And add the integration test in a second one inheriting from the first: https://docs.docker.com/compose/extends/

hfjn (approved):

PR #14, opened 2019-05-28
Add apply config command topics

Bibob7:


Pipfile
```diff
@@ -7,6 +7,7 @@ verify_ssl = true
 kafkaesque = {path = ".",editable = true,extras = ["test", "dev"]}
 
 [packages]
+pyyaml = "*"
```
hfjn: Please move this requirement to the `setup.py`. Otherwise it won't be installed as a part of esque.

esque/cli/commands.py
```diff
@@ -90,6 +92,19 @@ def create_topic(state: State, topic_name):
         TopicController(state.cluster).create_topic(topic_name)
 
 
+@esque.command("apply", help="Apply a configuration")
+@click.option(
+    "-f", "--file", help="Config file path", default=False
```
hfjn: `default=False` would return a `bool` False at this point. Which will lead to a fail in 102. Maybe you wanted to used `required`?

If it's not required you'd need to handle the case without passing that option though.

esque/cli/commands.py
```diff
@@ -90,6 +92,19 @@ def create_topic(state: State, topic_name):
...
+)
+@pass_state
+def apply(state: State, file):
+    if ensure_approval("Are you sure?", no_verify=state.no_verify):
+        yaml_data = yaml.load(open(file), Loader=yaml.Loader)
```
hfjn: Please use `safe_load` instead of `load`. Using plain load is a security issue since it enables arbitrary code execution. Also I don't think you necessarily need to pass the loade here.

esque/cli/commands.py
```diff
@@ -90,6 +92,19 @@ def create_topic(state: State, topic_name):
         TopicController(state.cluster).create_topic(topic_name)
 
 
+@esque.command("apply", help="Apply a configuration")
+@click.option(
+    "-f", "--file", help="Config file path", default=False
+)
+@pass_state
+def apply(state: State, file):
```
hfjn: typing for file is missing.

esque/cli/output.py
```diff
@@ -74,6 +72,17 @@ def pretty_dict(d: MutableMapping[str, Any]) -> str:
     return "\n".join(lines)
 
 
+def get_value(unit, value):
```
hfjn: please add typing for file and return value.

esque/topic.py
```diff
@@ -14,6 +13,24 @@
     unpack_confluent_config,
 )
 
+class TopicConfig:
```
hfjn: I think you could merge this class and Topic into one. 
Workflow would be like this: Create new `Topic` instance -> Pass Instance to `TopicController` -> Controller creates Topic and enriches instance with more information.

esque/topic.py
```diff
@@ -14,6 +13,24 @@
...
+    def __init__(self, name: str, num_partitions: int, replication_factor: int, config: Dict[str, str] = None):
+        self.name = name
+        self.num_partitions = num_partitions
+        self.replication_factor = replication_factor
+        self.config = config
+
+    def get_new_topic(self):
```
hfjn: Vendor objects should not be exposed and only used internally to ease eventual replacement of libraries. Just keep the new_topic as an instance variable and use the parent object during topic creation instead of `NewTopic`.

esque/topic.py
```diff
@@ -135,11 +152,52 @@ def create_topic(
...
+    def create_topics(self, topics: List[NewTopic]):
+        if not topics:
+            return
+        for new_topic in topics:
+            future_list = self.cluster.confluent_client.create_topics([new_topic])
+            ensure_kafka_futures_done(list(future_list.values()))
+
+    @raise_for_kafka_exception
+    @invalidate_cache_after
+    def alter_configs(self, configs: List[ConfigResource]):
```
hfjn: Where is that ConfigResource coming from? I assume it's an internal confluent object?
Would probably nicer to merge that with Topic and TopicConfig as well.

esque/topic.py
```diff
@@ -149,3 +207,34 @@ def delete_topic(self, topic_name):
...
+        config_resources = []
+        topics_diff = {}
+        for topic_config in editable_topics:
+            if not topic_config.config:
+                continue
+
+            conf = self.cluster.retrieve_config(ConfigResource.Type.TOPIC, topic_config.name)
+            config_list = unpack_confluent_config(conf)
+            new_config = topic_config.config
+            config_diff = {}
```
hfjn:
```
config_diff = {
	name: [str(value), str(new_config.get(name))]
	for name, value in config_list.items()
	if new_config.get(name) and str(new_config.get(name) != str(value))
}
```
🙂

hfjn (changes requested):
Please also check CI and run `black` and `flake8`.
Thanks for your contribution. 👍

hfjn:
Ah and we'll definetly need tests for everything that actually changes something about the cluster state.

esque/cli/commands.py
Bibob7: I got a deprecation warning, if I not pass the loader. This warning was the reason to add the loader here. However, with safe_load there is no deprecation warning anymore.

esque/topic.py
Bibob7: This is nice XD

esque/topic.py
Bibob7: But I don't have a cluster object in the command methods. I was thinking about two different constructors for the Topic class, but this is obviously not possible in Python.

hfjn: It's a bit hidden but you do. If you use `@pass_state` the state object is passed to your command which lazily creates a cluster object: https://github.com/real-digital/esque/blob/master/esque/cli/options.py#L31

Bibob7 (commented):
Please also check CI and run black and flake8. -> Done

esque/topic.py
```diff
@@ -15,12 +14,24 @@
 )
 
 
-class Topic:
-    def __init__(self, name: str, cluster: Cluster):
+class _Topic:
```
hfjn: Don't think this needs to be private.

hfjn (commented):
LGTM. Just one small thing.

hfjn (approved):

PR #15, opened 2019-06-11
Fix ping command

Bibob7:
The topic creation and deletion during the ping command was broken.
Instead of passing a Topic object only the topic name was passed.


MrTrustworthy:
I'd approve, but black is still failing for some reason:
`would reformat /home/travis/build/real-digital/esque/esque/cli/commands.py`

hfjn (commented):
Yup. Please fix CI so we can approve

hfjn (approved):

hfjn:
LGTM. Merge away.

PR #19, opened 2019-07-03
In-place edit

hfjn:
Adds basic in-place edit functionality for topics.
Calling `esque edit topic <topic_name>` opens an editor window based on the user's pre-defined editor and then edit the same configs which can be set via `apply`.

Also comes with basic `as_yaml()` method which should enable to export topics lists for `apply` in the future.


MrTrustworthy (approved):
LGTM

PR #20, opened 2019-07-03
Split test env setup and test running in docker compose files

### LINKED ISSUE #12: Create a base-docker-compose.yml with the stack.
We could create a base `docker-compose.yml` with the stack. And add the integration test in a second one inheriting from the first: https://docs.docker.com/compose/extends/

_Originally posted by @hfjn in https://github.com/real-digital/esque/pull/10_

MrTrustworthy:
Closes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,9 @@ topics:
 ```
 
 ## Development
+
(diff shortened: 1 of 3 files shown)


hfjn (approved):
LGTM

PR #21, opened 2019-07-05
Adds automatic autocompletion installation

### LINKED ISSUE #18: Autocompletion Initialization significantly slowing down shell
I load the `esque` autocompletion in my `.zshrc` this significantly raises the startup time of my shell.

To circumvent that we should generate the autocompletion once on install of the tool, or alternative see it as part of the package and ship it with the code.

Better way of generating the autocompletion according to the click documentation:
https://click.palletsprojects.com/en/7.x/bashcomplete/#activation-script

```
_ESQUE_COMPLETE=source_zsh esque > esque-complete.zsh

source esque-complete.zsh
```

hfjn:
Fixes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,15 @@ If it doesn't work for some reason you can still install it yourself:
 #### Bash
 
 ```
- echo 'eval "$(_ESQUE_COMPLETE=source esque)"' >> ~/.esque-autocompletion.sh
- echo "source ~/.esque-autocompletion.sh" >> ~/.bashrc
+ echo 'eval "$(_ESQUE_COMPLETE=source esque)"' >> ~/.esque/autocompletion.sh
+ echo "source ~/.esque/autocompletion.sh" >> ~/.bashrc
 ```
 
 #### ZSH
 
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -29,22 +29,6 @@ This is partly (but not only) due to a fragmented and unclear definition of tool
 
 `esque` is available at pypi.org and can be installed with `pip install esque`. `esque` requires Python 3.6+ to run.
 
-### Enable Autocompletion
```
MrTrustworthy, line 32: Can we still add a comment about autocompletion being installed automatically?

setup.py
```diff
@@ -31,6 +33,16 @@
     "pyyaml",
 ]
 
+
+class PostInstallCommand(install):
```
MrTrustworthy: `InstallWithPostCommand` or something would make it clearer that this is actually _also_ running the normal installation

MrTrustworthy (changes requested):

MrTrustworthy (approved):

swenzel (approved):

PR #27, opened 2019-07-07
Refactoring of topic handling in preparation of #22

### LINKED ISSUE #22: As a user, i want to dump all topic setting related information to disk, modify this, and then re-apply the changes

### LINKED ISSUE #26: `apply` returns OK even if topic wasn't created
Kafka logs may show a `org.apache.kafka.common.errors.InvalidReplicationFactorException: Replication factor: 3 larger than available brokers: 1`, yet esque will report `Successfully created topics: ...`

MrTrustworthy:
Fixes absolutely Nothing ~~#26~~


esque/broker.py
```diff
@@ -1,13 +1,46 @@
...
+    def __init__(
+        self, cluster, *, broker_id: int = None, host: str = None, port: int = None
+    ):
         self.cluster = cluster
         self.broker_id = broker_id
+        self.host = host
+        self.port = port
+
+    @classmethod
+    def from_id(cls, cluster, broker_id) -> "Broker":
```
swenzel, line 17: The constructor supports the exact same signature, so what's the benefit of this method?

esque/broker.py
```diff
@@ -1,13 +1,46 @@
...
+        return cls(cluster=cluster, broker_id=broker_id)
+
+    @classmethod
+    def from_attributes(cls, cluster, broker_id: int, host: str, port: int) -> "Broker":
```
swenzel, line 21: see above

esque/cli/commands.py
```diff
@@ -110,33 +97,50 @@ def create_topic(state: State, topic_name: str):
...
         )
 
 
+@edit.command("topic")
+@click.argument("topic-name", required=True)
+@pass_state
+def edit_topic(state: State, topic_name: str):
+    controller = TopicController(state.cluster)
+    topic = TopicController(state.cluster).get_cluster_topic(topic_name)
+    new_conf = click.edit(topic.to_yaml())
```
swenzel: Could you pass `extension='.yml'` to `click.edit`? I'd appreciate the syntax highlighting in the editor :smile:

esque/cli/commands.py
```diff
@@ -233,7 +227,7 @@ def get_offsets(state, topic_name):
 @click.argument("broker-id", required=True)
 @pass_state
 def describe_broker(state, broker_id):
-    broker = Broker(state.cluster, broker_id).describe()
+    broker = Broker.from_id(state.cluster, broker_id).describe()
```
swenzel, line 231: I'd actually prefer `state.cluster.get_broker(broker_id).describe()`

esque/cli/commands.py
```diff
@@ -258,9 +252,9 @@ def describe_consumergroup(state, consumer_id, verbose):
 @get.command("brokers")
 @pass_state
 def get_brokers(state):
-    brokers = state.cluster.brokers
+    brokers = Broker.get_all(state.cluster)
```
swenzel, line 252: Similar to above: `state.cluster.get_all_brokers()` would be nicer.
By the way, brokers can have multiple endpoints. Guess by default we should use the one that esque is using. But would be good to have the possibility to list all endpoints.

swenzel (commented):
Since its all over the place, I'll comment it here. I'd keep the `TopicController` in the `Cluster` object. It's best to have one single object (i.e. cluster) that you can use as interface.
The `TopicController` should exist only once per cluster anyway and who is better to ensure this if not the cluster object itself?
The cluster object could serve as a factory for all Kafka resources so you don't have to repeat yourself.
Then you can create a cluster with some bootstrap servers (and later also credentials) and everything else can (and should) be handled from there.

MrTrustworthy:
> Since its all over the place, I'll comment it here. I'd keep the TopicController in the Cluster object. It's best to have one single object (i.e. cluster) that you can use as interface.
> The TopicController should exist only once per cluster anyway and who is better to ensure this if not the cluster object itself?
> The cluster object could serve as a factory for all Kafka resources so you don't have to repeat yourself.
> Then you can create a cluster with some bootstrap servers (and later also credentials) and everything else can (and should) be handled from there.

I've also considered that, but decided against it in favour of the current approach. My thoughts behind that: I'd like to treat the Cluster as a "dumb" resource object similar to how the Topic should behave after this PR. The resource objects should only wrap the different clients and provide some basic utility functions. The TopicController just get's the Cluster as input, and can then create/change resources of that cluster as output. That's also how I'd structure further Controllers ex. for ConsumerGroups or Partitions.

I think this approach makes the View (the CLI) a little more verbose, but should make the separation between resources and controllers more explicit and hopefully testing/mocking easier as well.

But having the Cluster as god-object that manages all resources could also work, and I'm open to changing it to that approach if you think it would be better. We'll just have to agree on one architectural approach going forward.

esque/cli/commands.py
MrTrustworthy: s.A.

esque/cli/commands.py
MrTrustworthy: For the approach I'm working on, I'd use this as preparation to move this stuff to a BrokerController. See my comment at the bottom

esque/cli/commands.py
MrTrustworthy: done

esque/cli/commands.py
```diff
@@ -112,58 +113,56 @@ def edit_topic(state: State, topic_name: str):
...
+    if not ensure_approval("Apply changes?", no_verify=state.no_verify):
+        click.echo("Cancelling changes")
+        return
+
+    # apply changes
+    topic_controller.create_topics(to_create)
+    topic_controller.alter_configs(to_edit)
+
+    # output confirmation
+    changes = {"ignored": len(to_ignore), "created": len(to_create), "changed": len(to_edit)}
```
swenzel: "Unchanged" is probably better than "ignored", otherwise I'd ask myself why they got ignored.

tests/conftest.py
```diff
@@ -132,11 +133,15 @@ def confluent_admin_client(test_config: Config) -> AdminClient:
...
-        cluster.pykafka_client.cluster, topic_object._pykafka_topic, sync=True
-    )
+def producer(test_config: Config):
+    producer_config = test_config.create_confluent_config()
+    yield Producer(producer_config)
+
+@pytest.fixture()
+def avro_producer(test_config: Config):
+    producer_config = test_config.create_confluent_config()
+    producer_config.update({"schema.registry.url": Config().schema_registry})
```
swenzel: Could it be that you want `test_config.schema_registry` instead of  `Config().schema_registry` ?

MrTrustworthy: That's just in here because i merged in master :D

esque/cli/commands.py
MrTrustworthy: agreed, fixed

swenzel:
For the moment I'm okay with separating it. But personally I'd prefer a "god" object as you call it. It could be just a proxy. Imho its easier to have one object around where I can check the API using my IDE than to keep several where I first have to find and figure out how to create the one that does what I want. Looking from a usability perspective here.

MrTrustworthy:
Some of the newer changes also close #28 by simply not showing non-editable properties in `esque edit topic`. Still need a way to detect & abort/fail for `esque apply`

MrTrustworthy:
@swenzel If you want to do a follow-up PR to change it, go for it :)

swenzel (approved):

PR #30, opened 2019-07-08
Transfer command

Bibob7:


Bibob7:
Still todo: 

- Write tests for new logic.

esque/cli/commands.py
```diff
@@ -280,6 +296,130 @@ def get_topics(state, topic, o):
...
+    avro: bool,
+):
+    current_timestamp_milliseconds = int(round(time.time() * 1000))
+    temp_name = topic + "_" + str(current_timestamp_milliseconds)
+    group_id = "group_for_" + temp_name
+    directory_name = "temp-file_" + temp_name
+    base_dir = Path(directory_name)
+    state.config.context_switch(from_context)
+
+    with DeleteOnFinished(base_dir) as working_dir:
```
hfjn: I'm not sure if deletion of the temporary files should be expected behavior. At the very least we should give the possibility to keep those files.

esque/clients.py
```diff
@@ -40,22 +61,104 @@ def consume_ping(self) -> Optional[Tuple[str, int]]:
...
+
+        self._config.update(
+            {"default.topic.config": {"auto.offset.reset": offset_reset}}
+        )
+        self._consumer = confluent_kafka.Consumer(self._config)
+        self._assign_exact_partitions(topic_name)
+        self.file_writer = file_writer
+
+    @classmethod
+    def create(
```
hfjn: Please don't use factory methods.

esque/clients.py
```diff
@@ -40,22 +61,104 @@ def consume_ping(self) -> Optional[Tuple[str, int]]:
         delta_sent = pendulum.now() - msg_sent_at
         return msg.key(), delta_sent.microseconds / 1000
 
-    def _assign_exact_partitions(self, topic: str) -> None:
-        self._consumer.assign([TopicPartition(topic=topic, partition=0, offset=0)])
+
+class FileConsumer(Consumer):
+    def __init__(
```
hfjn: Was just thinking: I'd be cool if the basic consumer already had a `consume` method and the inheriting consumers just overwrite the processing of the message. I.e. -> Ping Consumer calculates the ping, File Consumer writes it to file.

esque/clients.py
```diff
@@ -40,22 +61,104 @@ def consume_ping(self) -> Optional[Tuple[str, int]]:
...
+        topic_name: str,
+        working_dir: pathlib.Path,
+        file_writer: FileWriter,
+        last: bool,
+    ):
+        super().__init__(group_id, topic_name, working_dir, file_writer, last)
+        self.writer = file_writer
+
+    @classmethod
+    def create(
```
hfjn: See above

esque/clients.py
```diff
@@ -64,3 +167,58 @@ def produce_ping(self):
...
+
+            while True:
+                left_messages = self._producer.flush(1)
+                if left_messages == 0:
+                    break
+                click.echo("Still {left_messages} messages left, flushing...")
+
+            return counter
+
+    def produce(self, topic_name: str, message: KafkaMessage):
```
hfjn: Same as for the consumer? Shared produce method and only creation of the kafka message in the inheriting producers.

esque/message.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class PlainTextFileWriter(FileWriter):
+    def write_message_to_file(self, message: Message, file: BinaryIO):
+        decoded_message = decode_message(message)
+        serializable_message = {
+            "key": decoded_message.key,
+            "value": decoded_message.value,
+        }
+        pickle.dump(serializable_message, file)
```
hfjn: Is the message "plaintext" after this? Could we just write it as json "alternatively"? Would be a nice feature if we decided to keep the temp folder.

hfjn (changes requested):
Please also run black with `--line-length 119`. Would be cool if you could add that to travis as well.

.travis.yml
```diff
@@ -4,12 +4,14 @@ cache: pip
 env:
 - TEST_CMD="pytest tests/ --integration --cov=esque --local"
 before_install:
```
swenzel, line 10: I'd suggest to swap lines 10 and 11 like here:
https://github.com/real-digital/pyconnect/blob/master/.travis.yml#L10
This way you'll wait for zookeeper before you start kafka which makes it start up quicker ;)

esque/avromessage.py
```diff
@@ -0,0 +1,103 @@
...
+import fastavro
+from confluent_kafka.cimpl import Message
+from confluent_kafka.avro import loads as load_schema
+
+from esque.message import FileWriter, FileReader, KafkaMessage
+from esque.schemaregistry import SchemaRegistryClient
+
+
+class DecodedAvroMessage:
+    def __init__(self, key: Optional[Dict], value: Optional[Dict], key_schema_id: int, value_schema_id: int):
```
swenzel: Theoretically key and value don't have to be dicts. A single primitive field can also be a valid avro message.

esque/avromessage.py
```diff
@@ -0,0 +1,103 @@
...
+
+import fastavro
+from confluent_kafka.cimpl import Message
+from confluent_kafka.avro import loads as load_schema
+
+from esque.message import FileWriter, FileReader, KafkaMessage
+from esque.schemaregistry import SchemaRegistryClient
+
+
+class DecodedAvroMessage:
```
swenzel: This could actually be a `NamedTuple` (https://docs.python.org/3.6/library/typing.html#typing.NamedTuple).

esque/avromessage.py
```diff
@@ -0,0 +1,103 @@
...
+            schema_directory = self.directory / record["schema_directory_name"]
+
+            key_schema = load_schema((schema_directory / "key_schema.avsc").read_text(encoding="utf-8"))
+            value_schema = load_schema((schema_directory / "value_schema.avsc").read_text(encoding="utf-8"))
+
+            yield KafkaMessage(record["key"], record["value"], key_schema, value_schema)
+
+
+def extract_schema_id(message: bytes) -> int:
+    _, schema_id = struct.unpack(">bI", message[:5])
```
swenzel: It's probably better to not ignore the first byte. I'd do it like this:
```python
magic_byte, schema_id = struct.unpack(">bI", message[:5])
assert magic_byte == 0, f"Wrong magic byte ({magic_byte}), no AVRO message."
```

esque/clients.py
```diff
@@ -40,27 +75,105 @@ def consume_ping(self) -> Optional[Tuple[str, int]]:
...
+    def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, last: bool):
+        super().__init__(group_id, topic_name, last)
+        self.working_dir = working_dir
+        offset_reset = "earliest"
+        if last:
+            offset_reset = "latest"
+
+        self._config.update({"default.topic.config": {"auto.offset.reset": offset_reset}})
+        self._consumer = confluent_kafka.Consumer(self._config)
+        self._subscribe(topic_name)
```
swenzel, line 77: Believe it or not, there was a reason why I didn't support multiple partitions in my script :D you can't just ignore it...
Consider for example non-default partitioning where a producer decides for itself which partition to produce the message to.
Also in case of AVRO, the schema id can change at different offsets within the partitions. Your `AvroFileWriter` can't handle that.
I think the easiest solution would be to create a separate file writer for every partition and let it write into its own subfolder based on that partition. This way you don't have to change the logic within `AvroFileWriter`. In the file writer I'd still make it clear, however, that it can only deal with messages from a single partition. Either by documenting it or by checking that `message.partition()` within `write_message_to_file` always returns the same number.

esque/schemaregistry.py
```diff
@@ -0,0 +1,25 @@
...
+SchemaPair = namedtuple("SchemaPair", ["original_schema", "parsed_schema"])
+
+
+class SchemaRegistryClient:
+    def __init__(self, schema_registry_uri: str):
+        self.schema_registry_uri = schema_registry_uri
+
+    @lru_cache(maxsize=100)
+    def get_schema_from_id(self, schema_id: int) -> SchemaPair:
+        return self.get_schema_from_server(schema_id, self.schema_registry_uri)
```
swenzel: Please merge the two methods. The other one was meant as a function but doesn't make sense as a method for an object that already knows the server url.

tests/conftest.py
```diff
@@ -156,14 +153,35 @@ def consumer(topic_object: Topic, consumer_group):
 @pytest.fixture()
 def filled_topic(producer, topic_object):
```
swenzel, line 177: Maybe we can turn the `topic_object` into a topic factory. Then we could create topics with multiple partitions.
You can also have a look at: https://github.com/real-digital/pyconnect/blob/master/test/conftest.py#L112

tests/integration/test_clients.py
```diff
@@ -0,0 +1,62 @@
+import pathlib
```
swenzel, line 2: None of your tests are checking for content. How do you know the messages are consumed/produced correctly?

tests/test_samples/key_schema
```diff
@@ -0,0 +1,8 @@
+{
```
swenzel: The commonly accepted file ending for avro schemas is '.avsc'.

swenzel (changes requested):

esque/clients.py
Bibob7: > Consider for example non-default partitioning where a producer decides for itself which partition to produce the message to.
Also in case of AVRO, the schema id can change at different offsets within the partitions. Your AvroFileWriter can't handle that.

Actually it does, if I understood you correctly. I store for each serialized message the related schema_directory. And for each message it the schema ids.

esque/clients.py
```diff
@@ -88,27 +90,38 @@ def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, la
...
-        with self.file_writer:
+        file_writers = {}
+        with ExitStack() as stack:
             while counter < amount:
                 message = self._consume_single_message()
                 if message is None:
                     return counter
 
-                self.file_writer.write_message_to_file(message)
+                if message.partition() not in file_writers.keys():
```
swenzel: Membership tests and iteration over dictionaries go over keys by default. So you actually don't need to add `.keys()`.

esque/clients.py
```diff
@@ -88,27 +90,38 @@ def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, la
...
+                    file_writer = self.get_file_writer((self.working_dir / f"partition_{message.partition()}"))
```
swenzel: The FileConsumer already knows its directory. So you don't have to pass it here. I'd call the method `get_file_writer_for_partition` and just pass the partition.
Also since you call `message.partition()` quite often it could make sense to assign that to a local variable.

esque/clients.py
```diff
@@ -142,33 +155,41 @@ def __init__(self, working_dir: pathlib.Path):
...
 
     def produce(self, topic_name: str) -> int:
-        with self.file_reader:
-            counter = 0
-            for message in self.file_reader.read_from_file():
-                self.produce_message(topic_name, message)
-                counter += 1
+        path_list = glob(str(self.working_dir / "partition_*"))
+        counter = 0
+        with ExitStack() as stack:
```
swenzel: Since you're processing the files sequentially, you don't actually need the exit stack here. You could do `with self.get_file_reader(pathlib.Path(partition_path)) as file_reader:` within the for loop

.travis.yml
```diff
@@ -26,11 +32,13 @@ notifications:
 matrix:
   include:
   - name: black
+    before_install: []
     python: 3.7
     env:
-    - TEST_CMD="black --check --verbose ."
+    - TEST_CMD="black --check --verbose --line-length=119 ."
```
swenzel: Line length now is in pyproject.toml

README.md
```diff
@@ -134,6 +134,17 @@ While this `docker-compose` stack is up, you can run the tests from the CLI via
...
 
 
+### Pre Commit Hooks
+
+To install pre commit hooks run:
+
+```
+pip install pre-commit
+pre-commit install
+pre-commit install-hooks`
```
swenzel: There is one back tick too much

esque/cli/commands.py
```diff
@@ -274,12 +227,85 @@ def get_consumergroups(state):
...
     for topic in topics:
         click.echo(topic.name)
 
 
+@esque.command("transfer", help="Transfer messages of a topic from one environment to another.")
+@click.argument("topic", required=True)
+@click.option("-f", "--from", "from_context", help="Source Context", type=click.STRING, required=True)
+@click.option("-t", "--to", "to_context", help="Destination context", type=click.STRING, required=True)
+@click.option("-n", "--numbers", help="Number of messages", type=click.INT, required=True)
+@click.option("--last/--first", default=False)
```
swenzel: Can you provide a help message for this parameter?

esque/clients.py
```diff
@@ -1,37 +1,75 @@
...
+    def _consume_single_message(self) -> Optional[Message]:
+        poll_limit = 10
+        counter = 0
+        while counter < poll_limit:
+            message = self._consumer.poll(timeout=10)
+            if message is None:
+                counter += 1
+                continue
+            if message.error() is not None:
+                if message.error().code() == KafkaError._PARTITION_EOF:
```
swenzel: `enable.partition.eof` is set to `False` by default (see constructor above) so usually this will be False all the time. Also this only works if you have a single partition. In case of multiple partitions, there could still be a message in another partition, so you don't want to break out of the loop.
However, I'm not quite sure how this method is supposed to behave at EOF.  

What exactly is all this logic used for anyway?
From what I can see, it is reimplementing the retry logic that is already present in `poll`. So it could just be a very simple wrapper around it:
```python
def _consume_single_message(self, timeout=10) -> Optional[Message]:
    message = self._consumer.poll(timeout=timeout)
    raise_for_message(message)
    return message
```

esque/clients.py
```diff
@@ -1,37 +1,75 @@
...
+                else:
+                    raise RuntimeError(message.error().str())
+            raise_for_message(message)
+            return message
+
 
-    def consume_ping(self) -> Optional[Tuple[str, int]]:
+class PingConsumer(AbstractConsumer):
+    def consume(self, amount: int) -> Optional[Tuple[str, int]]:
         msg = self._consumer.consume(timeout=10)[0]
```
swenzel: You should use `self._consume_single_message` here. Then you can also get rid of `raise_for_message(msg)` which is already called there.

esque/message.py
```diff
@@ -0,0 +1,81 @@
...
+        pass
+
+
+class FileReader(IOHandler):
+    def read_from_file(self) -> Iterable[KafkaMessage]:
+        pass
+
+
+class PlainTextFileWriter(FileWriter):
+    def __init__(self, directory: pathlib.Path):
```
swenzel: You could move this whole constructor to `FileWriter`, I guess all writers need that mode.

esque/message.py
```diff
@@ -0,0 +1,81 @@
...
+        serializable_message = {
+            "key": decoded_message.key,
+            "value": decoded_message.value,
+            "partition": decoded_message.partition,
+        }
+        self.file.write(json.dumps(serializable_message) + "\n")
+
+
+class PlainTextFileReader(FileReader):
+    def __init__(self, directory: pathlib.Path):
```
swenzel: See above

esque/message.py
```diff
@@ -0,0 +1,81 @@
...
+class IOHandler:
+    def __init__(self, directory: pathlib.Path):
+        self.directory = directory
+        self.file_name = "data"
+        self.open_mode = "w+"
+
+    def __enter__(self):
+        if not self.directory.exists():
+            self.directory.mkdir()
+        self.file = (self.directory / self.file_name).open(self.open_mode)
```
swenzel, line 31: I'd return `self` here, so you can do `with IOHandler('somedir') as handler:`

esque/schemaregistry.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class SchemaRegistryClient:
+    def __init__(self, schema_registry_uri: str):
+        self.schema_registry_uri = schema_registry_uri
+
+    @lru_cache(maxsize=100)
+    def get_schema_from_id(self, schema_id: int) -> SchemaPair:
+        url = f"{self.schema_registry_uri}/schemas/ids/{schema_id}"
+        response = requests.get(url.format(**locals()))
```
swenzel: `url` is already formatted :smile:

esque/cli/commands.py
Bibob7: This is already provided by click:
`pbs $ esque transfer --help
Usage: esque transfer [OPTIONS] TOPIC

  Transfer messages of a topic from one environment to another.

Options:
  -f, --from TEXT        Source Context  [required]
  -t, --to TEXT          Destination context  [required]
  -n, --numbers INTEGER  Number of messages  [required]
  --last / --first
  -a, --avro             Set this flag if the topic contains avro data
  -k, --keep             Set this flag if the file with consumed messages
                         should be kept.
  --help                 Show this message and exit.`

esque/clients.py
Bibob7: There is a retry logic in poll? If this is the case, you are right. I will change it. Regarding EOF, you are also right. :)

esque/cli/commands.py
swenzel: I specifically meant the last/first Parameter

swenzel (approved):

esque/cli/commands.py
```diff
@@ -274,12 +227,85 @@ def get_consumergroups(state):
...
+    group_id = "group_for_" + unique_name
+    directory_name = "message_" + unique_name
+    base_dir = Path(directory_name)
+    state.config.context_switch(from_context)
+
+    with HandleFileOnFinished(base_dir, keep_file) as working_dir:
+        number_consumed_messages = _consume_to_file(working_dir, topic, group_id, from_context, numbers, avro, last)
+
+        if number_consumed_messages == 0:
+            click.echo(click.style("Execution stopped, because no messages consumed.", fg="red"))
```
hfjn, line 266: I would state the reason for stopping the execution. No consumption is the effect. This case is probably getting hit when the topic is empty or the starting offset was chosen to high, right?

hfjn (approved):
One minor comment. Otherwise approved.
Nice work and great code. 👍

esque/cli/commands.py
Bibob7: How would you do this? Maybe checking the high_watermarks for all topics as indicator for an empty topic? But with log retention it could also be empty with a high_watermark greater 0.

hfjn: I didn't mean in that much detail. 
The user can still find out what exactly the problem was by him/herself. You could just name the possible reasons in the error message.

Bibob7: ah sure

PR #31, opened 2019-07-11
Use milliseconds per year as threshold

Bibob7:
Use milliseconds per year as threshold for prettifying output.


esque/cli/output.py
```diff
@@ -117,7 +117,7 @@ def pretty_duration(value: Any, *, multiplier: int = 1) -> str:
     value *= multiplier
 
     # Fix for conversion errors of ms > C_MAX_INT in some internal lib
-    if value > C_MAX_INT:
+    if value > MILLISECONDS_PER_YEAR:
         value = int(value / 1000 / 3600 / 24 / 365)
```
swenzel: You could actually do `int(value / MILLISECONDS_PER_YEAR)` here

hfjn (approved):
LGTM

PR #39, opened 2019-09-04
Fix json problem with avro-python3 1.9 version

Bibob7:
With avro-python3 version 1.9 I had the same problem as described here: https://github.com/confluentinc/confluent-kafka-python/issues/610

When producing avro messages, it always wanted to json encode MappingProxys, which is not possible. Downgrading to version 1.8.2 solved the problem.

I also added a produce command, which helped me to test my "solution" and which could also be helpful for others.


esque/message.py
```diff
@@ -15,8 +16,8 @@ class KafkaMessage(NamedTuple):
     key: Any
     value: Any
     partition: int
-    key_schema: str = None
-    value_schema: str = None
+    key_schema: RecordSchema = None
```
Bibob7: Based on the logic this is always a RecordSchema. str was actually wrong.

esque/cli/commands.py
```diff
@@ -311,10 +311,21 @@ def transfer(
...
 
         state.config.context_switch(to_context)
-        _produce_from_file(topic, to_context, working_dir, avro)
+        _produce_from_files(topic, to_context, working_dir, avro)
 
 
-def _produce_from_file(topic: str, to_context: str, working_dir: pathlib.Path, avro: bool):
+@esque.command("produce", help="Produce messages from directory based on output from transfer command")
+@click.argument("topic", required=True)
+@click.option(
```
hfjn, line 337: Now that I think about it: In my opinion it is unreasonable to have required options. We should probable either make it an argument or set a default. E.g. fall back to the current directory if nothing is given. What do you think @swenzel ?

esque/cli/commands.py
```diff
@@ -311,10 +311,21 @@ def transfer(
...
+    "-d", "--directory", help="Sets the directory that contains Kafka messages", type=click.STRING, required=True
```
hfjn: I would prefer to hand in a path instead of a directory. You could use the pathlib's globbing feature to hand in folders by passing `folder/*.avro`. Still would give us the possibility to only hand in a file.

esque/cli/commands.py
```diff
@@ -311,10 +311,21 @@ def transfer(
...
+)
+@click.option("-a", "--avro", help="Set this flag if the topic contains avro data", default=False, is_flag=True)
```
hfjn, line 345: What happens if I forget to set this flag but the data is avro? Wouldn't it be possible to detect avro or not automatically based on the first few lines?

hfjn (changes requested):

esque/cli/commands.py
Bibob7: I also thought about fallback to current directory.

esque/cli/commands.py
Bibob7: This is definitely better then using this flag. But I would change this in another branch.

hfjn: Fair. Created an issue for it: https://github.com/real-digital/esque/issues/41

esque/cli/commands.py
swenzel: Well `kapp` also has required options... They're a little awkward since you don't expect them and it's not very obvious from the command description that they're required.  
Therefore I'd rather do it like 'docker build': Use an argument which requires a path. Even when it's `.`.

Also please don't forget to describe the arguments in the command help parameter. That works a little different from describing options. Some stupid example:
```python
@click.command(help=f"""
This command does <action> on <directory>
Where <action> may be one of:
 - dothis: for doing this
 - dothat: for doing that

And <directory> is a directory full of stuff.
""")
@click.argument("action", metavar="<action>")
@click.argument("directory", metavar="<directory>", type=pathlib.Path)
def do(action: str, directory: pathlib.Path):
  pass
```

esque/cli/commands.py
Bibob7: We will do this in https://github.com/real-digital/esque/issues/70

hfjn (approved):

PR #44, opened 2019-09-05
Find a consistent, non-surprising way to handle replication factor and partition amount for new and existing topics

### LINKED ISSUE #43: Find a consistent, non-surprising way to handle replication factor and partition amount for new and existing topics

MrTrustworthy:
Closes #43


MrTrustworthy:
This PR is a bigger than intended, since I've had to refactor the entire Diff logic to make a consistent check possible, and I ran in some unexpected difficulties (ex. the config values we get from the cluster are always strings, but for local topics we can provide values as either string or integers). 

We now have a TopicDiff class with some more utility functions to keep the validation logic etc. in one place. This should also make integrating #45 into `esque apply` a little easier to implement and test.

MrTrustworthy:
Weird failure in the travis tests that doesn't show up on local tests...

`<TopicDiff[{'replication_factor': <AttributeDiff[remote:0, local:1]>}>`

How the F is it possible to have a replication factor of 0 on a remote topic o_O

esque/cli/commands.py
```diff
@@ -172,11 +172,13 @@ def apply(state: State, file: str):
         click.echo("No changes detected, aborting")
         return
 
-    # Warn users when replication & num_partition changes are attempted
-    if len(to_edit) > 0:
+    # Warn users & abort when replication & num_partition changes are attempted
+    if any(not diff.is_valid for _, diff in to_edit_diffs.items()):
```
hfjn, line 176: Just to be clear. That is something we will tackle in upcoming issues, right? At least raising this should be possible imo.

scripts/init.sh
```diff
@@ -0,0 +1,34 @@
+function updateConfig() {
```
hfjn, line 3: Probably should add a comment with the wurstmeister kafka init script here. Just to give credit. https://github.com/wurstmeister/kafka-docker/blob/master/start-kafka.sh

hfjn (approved):
LGTM. Black and flake are broken though.

scripts/init.sh
hfjn: Done.

hfjn:
> LGTM. Black and flake are broken though.

Already ran black. Should be fine now.

esque/cli/commands.py
swenzel: Also the comment and first output know about logic that is hidden in `diff.is_valid`. Maybe it's better to have something called `diff.validate()` that will raise an exception with information about the reason why it is not valid. Otherwise `is_valid` will change and the comment and output here will stay the way they are and mislead you.

MrTrustworthy: yes, this would be the follow up after you get that particular feature stable :)

esque/topic.py
```diff
@@ -41,6 +40,87 @@ def as_dict(self):
...
+        # since all "config" variables we get from cluster topics are sent to us as strings,
+        # we need to convert the local ones to string to match them correctly (exclude None's)
+        local = str(local) if isinstance(remote, str) and local is not None else local
+
+        if remote == local:
+            return self
+
+        # TODO: this should be handled correctly by checking the cluster defaults, like
+        # if remote == cluster_default(name) and local is None: return
+        # currently, if an attribute that was set get's un-set, it's ignored
```
swenzel, line 87:
```suggestion
        # currently, if an attribute is set remotely but supposed to be unset now, that won't happen because it's ignored
```

esque/topic.py
```diff
@@ -41,6 +40,87 @@ def as_dict(self):
...
+        return self
+
+    @property
+    def is_valid(self) -> bool:
+        return set(self._diffs.get(a, None) for a in self.INVALID_CHANGES) == {None}
+
+    @classmethod
+    def from_dict(cls, diff_dict: Dict[str, AttributeDiff]) -> "TopicDiff":
+        td = TopicDiff()
+        td._diffs = diff_dict
```
swenzel, line 106:
```suggestion
        td._diffs = diff_dict.copy()
```

esque/topic.py
```diff
@@ -90,6 +157,20 @@ def offsets(self) -> Dict[int, Watermark]:
...
+        if self.is_only_local:
+            return self.__num_partitions
+        return len(self.partitions)
+
+    @property
+    def replication_factor(self) -> int:
+        if self.is_only_local:
+            return self.__replication_factor
+        partition_replication_factors = set(len(p.partition_replicas) for p in self.partitions)
+        assert len(partition_replication_factors) == 1, "Different replication factors for partitions!"
```
swenzel, line 171: That might actually happen during partition reassignment... We could either try to find the most occurring value or tell the user to try again later.

esque/topic_controller.py
```diff
@@ -1,26 +1,48 @@
...
     from esque.cluster import Cluster
 
-AttributeDiff = namedtuple("AttributeDiff", ["old", "new"])
+
+class ClientTypes(Enum):
+    Confluent = "Confluent"
+    PyKafka = "PyKafka"
+
+
+ClientType = Union[ConfluentTopic, PyKafkaTopic]
```
swenzel, line 23: Maybe `ClientTopicType`? This one is easy to confuse with `ClientTypes`.

esque/topic_controller.py
```diff
@@ -1,26 +1,48 @@
...
 
+    @raise_for_kafka_exception
+    def _get_client_topic(self, topic_name: str, client_type: ClientTypes) -> ClientType:
+        confluent_topics = self.cluster.confluent_client.list_topics(topic=topic_name, timeout=10).topics
+        if client_type == ClientTypes.Confluent:
+            return confluent_topics[topic_name]
+        elif client_type == ClientTypes.PyKafka:
+            # at least PyKafka does it's own caching, so we don't have to bother
+            pykafka_topics = self.cluster.pykafka_client.cluster.topics
+            # PyKafka will try to auto-create the topic if you use `topic[random_name]`
```
swenzel, line 39:
```suggestion
            # PyKafka will try to auto-create the topic if you use `topic[topic_name]`, so we check first if it exists in `confluent_topics`, we can't check if it exists in `pykafka_topics` since that one lazily evaluated and won't contain it.
```

swenzel (approved):

PR #47, opened 2019-09-12
fix and rename ensure_kafka_future_done, add operation_timeout to .cl…

garrettthomaskth:
…uster.confluent_client.create_topics


hfjn (approved):
Thx.

PR #52, opened 2019-09-17
Just moved some stuff around because it felt more logical

hfjn:
Just splitted some stuff, created a few modules and ran isort. Nothing special.


esque/controller/topic_controller.py
```diff
@@ -8,8 +8,8 @@
```
swenzel: The `ConsumerGroupController` is in `esque.resources.consumergroup`. So either move that to `esque.controllers` or move this to `esque.resources.topic` to stay consistent. I think I'd prefer the former solution since it's closter to the MVC pattern which we're kind of using :smile:

swenzel (changes requested):
Love it!

esque/controller/topic_controller.py
hfjn: Fair. Moved it.

swenzel (approved):

PR #54, opened 2019-09-18
Add more tests

hfjn:
Together with #53 this should cover almost all cli commands.


esque/clients.py
```diff
@@ -56,8 +57,32 @@ def _consume_single_message(self, timeout=30) -> Optional[Message]:
 
 
 class PingConsumer(AbstractConsumer):
+    def __init__(self, group_id: str, topic_name: str, last: bool):
```
garrettthomaskth: That is a lot of repeated lines. What about removing self._subscribe(topic_name) from line 44 in the AbstractConsumer class, and then just 
super().__init__(group_id, topic_name, last)
self._assign_exact_partitions(topic_name)

esque/clients.py
```diff
@@ -56,8 +57,32 @@ def _consume_single_message(self, timeout=30) -> Optional[Message]:
...
+            }
+        )
+        self._consumer = confluent_kafka.Consumer(self._config)
+        self._assign_exact_partitions(topic_name)
+
+    def _assign_exact_partitions(self, topic: str) -> None:
+        self._consumer.assign([TopicPartition(topic=topic, partition=0, offset=0)])
+
     def consume(self, amount: int) -> Optional[Tuple[str, int]]:
-        message = self._consume_single_message()
```
garrettthomaskth: unused but required parameter, the best thing I could think of to do with this was the insert statement.

tests/integration/commands/test_deletion.py
```diff
@@ -0,0 +1,52 @@
...
+    assert result.exit_code == 0
+
+    # Invalidate cache
+    confluent_admin_client.poll(timeout=1)
+    topics = confluent_admin_client.list_topics(timeout=5).topics.keys()
+    assert topic not in topics
+
+
+@pytest.mark.integration
+def test_keep_kafka_duplicated(
```
garrettthomaskth: I think this name is a little confusing. What about 'test_keep_minus_delete_period' or something like that?

tests/integration/commands/test_deletion.py
```diff
@@ -0,0 +1,52 @@
...
+
+from esque.cli.commands import delete_topic
+
+
+@pytest.fixture()
+def basic_topic(num_partitions, topic_factory):
+    yield from topic_factory(num_partitions, "basic-topic")
+
+
+@pytest.fixture()
```
garrettthomaskth, line 13: There are also collisions with underscore and period, we could test that too 
https://jaceklaskowski.gitbooks.io/apache-kafka/kafka-topics.html

tests/integration/commands/test_ping.py
```diff
@@ -0,0 +1,25 @@
...
+from esque.topic_controller import TopicController
+
+
+def test_smoke_test_ping(cli_runner: CliRunner):
+    result = cli_runner.invoke(ping)
+
+    assert result.exit_code == 0
+
+
+def test_correct_amount_of_messages(
```
garrettthomaskth: This doesn't actually test the amount of messages

tests/integration/test_topic_controller.py
```diff
@@ -139,144 +121,3 @@ def test_topic_diff(topic_controller: TopicController, topic_id: str):
...
-    apply_conf = {"topics": [topic_1]}
-
-    # having the same topic name twice in apply should raise an ValueError
-    path = save_yaml(topic_id, apply_conf)
-    result = runner.invoke(apply, ["-f", path], input="Y\n")
-    assert result.exit_code != 0 and isinstance(result.exception, KafkaException), f"Calling apply should have failed"
-
-
-@pytest.mark.integration
-def test_topic_creation_with_template_works(
```
garrettthomaskth, line 249: what happened to this test? :(

tests/integration/test_topic_controller.py
```diff
@@ -1,13 +1,8 @@
 import json
-from typing import Dict, Any
 
 import confluent_kafka
 import pytest
-import yaml
-from click.testing import CliRunner
 
-from esque.cli.commands import apply, create_topic
-from esque.cli.options import State
```
garrettthomaskth, line 10: This is deleted but there is still a call to State() in line 13

tests/integration/commands/test_deletion.py
hfjn: Yeah I know. Tried that. But kafka doesn't even let me create two colliding topics.

tests/integration/commands/test_ping.py
hfjn: Hm. Apparently I stopped halfway through. :D

esque/clients.py
hfjn: Yeah. Maybe we should move that that amoun argument to keywords or something. I don't really see the need for it to be in the abstract consumer.

esque/clients.py
hfjn: 👍

esque/clients/consumer.py
```diff
@@ -51,13 +51,19 @@ def _consume_single_message(self, timeout=30) -> Optional[Message]:
```
garrettthomaskth, line 51: Also because raise_for_message raises an exception when the message is none the output can be just Message, not Optional[Message]

esque/clients/consumer.py
```diff
@@ -35,29 +35,35 @@ def __init__(self, group_id: str, topic_name: str, last: bool):
...
 
 
 class PingConsumer(AbstractConsumer):
-    def consume(self, amount: int) -> Optional[Tuple[str, int]]:
-        message = self._consume_single_message()
+    def __init__(self, group_id: str, topic_name: str, last: bool):
+        super().__init__(group_id, topic_name, last)
+        self._assign_exact_partitions(topic_name)
 
+    def consume(self) -> Optional[Tuple[str, int]]:
```
garrettthomaskth, line 58: This also shouldn't be optional

garrettthomaskth (commented):
everything looks good to me :) just that the output of consume can't be None

swenzel (approved):

PR #64, opened 2019-09-23
Initial draft of the message filtering module.

ognjen-j:


esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,211 @@
...
+        elif self.generic_operator.name == Operator.OPERATORS["LOGICAL_AND"].name:
+            if boolean_operands:
+                return op1_converted and op2_converted
+            else:
+                raise TypeError("Logical operators can only be used with boolean values")
+        elif self.generic_operator.name == Operator.OPERATORS["LOGICAL_OR"].name:
+            if boolean_operands:
+                return op1_converted or op2_converted
+            else:
+                raise TypeError("Logical operators can only be used with boolean values")
```
hfjn: With all those TypeErrors. I think I would introduce some validator like `raise_for_operator` which checks if the operator fits the arguments once. That should keep your code DRY.

esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,211 @@
...
+            # maybe they are date strings
+            op1_converted = h.to_date_time(operand1)
+            op2_converted = h.to_date_time(operand2)
+        else:
+            # otherwise, we'll test them as strings
+            op1_converted = operand1
+            op2_converted = operand2
+        if op1_converted is None or op2_converted is None:
+            raise TypeError("The operands (" + operand1 + ", " + operand2 + ") cannot be compared.")
+        if self.generic_operator.name == Operator.OPERATORS["GREATER_THAN"].name:
```
hfjn: Why not map those operator types with a dict?

esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+class ComparisonBinaryOperator(AbstractBinaryOperator):
+    def evaluate(self, operand1: str, operand2: str):
+        op1_converted = None
+        op2_converted = None
+        if h.is_boolean(operand1) and h.is_boolean(operand2):
+            op1_converted = h.to_boolean(operand1)
+            op2_converted = h.to_boolean(operand2)
+        if h.is_float(operand1) and h.is_float(operand2):
```
hfjn: elif?

esque/ruleparser/fieldeval.py
```diff
@@ -0,0 +1,59 @@
...
+        self.__message = message
+        self.__header_pattern = re.compile(Operator.FIELDS["MESSAGE_HEADER"], flags=re.IGNORECASE)
+
+    def evaluate_field(self, field_name: str):
+        """ This method contains the logic to evaluate the field values. New fields are added as follows:
+            (1) add a new element to the Operator.FIELDS set. The key is arbitrary, and the field contains a regex for the parser,
+            (2) add an implementation to this method
+        """
+        if field_name == Operator.FIELDS["SYSTEM_TIMESTAMP"].replace("\\",""):
+            return datetime.datetime.strftime(datetime.datetime.now(), self.__time_format)
```
hfjn, line 24:
```suggestion
            return datetime.datetime.strftime(datetime.datetime.now(), self.__time_format)
         elif if self.__message is None:
              return -1
```

esque/ruleparser/fieldeval.py
```diff
@@ -0,0 +1,59 @@
...
+        elif field_name == Operator.FIELDS["MESSAGE_OFFSET"].replace("\\",""):
+            if self.__message is None:
+                return -1
+            return self.__message.offset()
+        elif field_name == Operator.FIELDS["MESSAGE_PARTITION"].replace("\\",""):
+            if self.__message is None:
+                return -1
+            partition = self.__message.partition()
+            if partition is None:
```
hfjn:
```suggestion
            if partition:
                return partition
            return -1
```

esque/ruleparser/fieldeval.py
```diff
@@ -0,0 +1,59 @@
...
+        elif field_name == Operator.FIELDS["MESSAGE_TIMESTAMP"].replace("\\",""):
+            if self.__message is None:
+                return -1
+            return datetime.datetime.strftime(datetime.datetime.fromtimestamp(self.__message.timestamp()[1]/1000.0), self.__time_format)
+        elif field_name == Operator.FIELDS["MESSAGE_LENGTH"].replace("\\",""):
+            if self.__message is None:
+                return -1
+            return len(self.__message)
+        elif field_name == Operator.FIELDS["MESSAGE_KEY"].replace("\\",""):
+            print(self.__message.key().decode("UTF-8"))
```
hfjn: maybe remove that print?

hfjn (changes requested):
Great work! Maybe try to modularize a bit more. You have lots of duplicated and nested code which makes it hard to read.

esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,220 @@
...
+        elif self.generic_operator.name == Operator.OPERATORS["LOGICAL_XOR"].name:
+            if boolean_operands:
+                return op1_converted ^ op2_converted
+            else:
+                raise TypeError("Logical operators can only be used with boolean values")
+
+
+class ComparisonBinaryOperator(AbstractBinaryOperator):
+    def evaluate(self, operand1: str, operand2: str):
+        op1_converted = None
```
garrettthomaskth: op1_converted = None
op2_converted = None
these don't do anything because there is a catch all else statement that assigns values to them

esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,220 @@
...
+            h.is_date_string(operand2) or h.is_date_time_string(operand2)
+        ):
+            # maybe they are date strings
+            op1_converted = h.to_date_time(operand1)
+            op2_converted = h.to_date_time(operand2)
+        else:
+            # otherwise, we'll test them as strings
+            op1_converted = operand1
+            op2_converted = operand2
+        if op1_converted is None or op2_converted is None:
```
garrettthomaskth: this can't happen

esque/ruleparser/expressionelement.py
garrettthomaskth: Would it be worth pulling out definitions into functions and then calling them from OPERATORS. And changing OPERATORS to have the name as a key, so we could just say 
Operator.OPERATORS[self.generic_operator.name].function(op1_converted, op2_converted)

esque/ruleparser/expressionelement.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class AbstractUnaryOperator(Operator):
+    def evaluate(self, operand1):
+        pass
+
+
+class ArithmeticBinaryOperator(AbstractBinaryOperator):
+    def evaluate(self, operand1: str, operand2: str):
+        op1_converted = None
```
garrettthomaskth: the first part of the evaluate function is the same for ArithmeticBinaryOperator and ComparisonBinaryOperator, how about pulling them out into a function in AbstractBinaryOperator called OperandTypeConversions or something?

esque/ruleparser/helpers.py
```diff
@@ -0,0 +1,119 @@
...
+    if input_string.lower() == "true":
+        return True
+    elif input_string.lower() == "false":
+        return False
+    else:
+        return None
+
+
+def to_date_time(input_string: str):
+    """This method will try to convert the string into a datetime object even if it is shorter than necessary.
```
garrettthomaskth: What about using dateutil.parser.parse. I think that function does the same thing

garrettthomaskth: then we don't have to test it :)

hfjn: Or use pendulum. it’s already a requirement anyway.

esque/ruleparser/expressionelement.py
ognjen-j: The major refactoring should have resolved these issues.

esque/ruleparser/expressionelement.py
ognjen-j: The major refactoring should have resolved these issues.

esque/ruleparser/fieldeval.py
ognjen-j: Fixed.

esque/ruleparser/expressionelement.py
ognjen-j: The major refactoring should have resolved these issues. The methods are, hopefully, much cleaner now.

esque/ruleparser/expressionelement.py
ognjen-j: Moved the evaluation to the tuple defining the operators.

esque/ruleparser/expressionelement.py
ognjen-j: Done

esque/ruleparser/fieldeval.py
ognjen-j: Done.

esque/ruleparser/expressionelement.py
ognjen-j: The refactoring should have solved this.

esque/ruleparser/helpers.py
ognjen-j: Pendulum already has the same functionality, so this issue resolved itself :)

hfjn (approved):
The refactoring really helped. :) Thanks!

hfjn:
Ah. And apparently you have merge conflicts.

PR #71, opened 2019-09-23
Updated the readme.md file according to the new vision document. TODO…

ognjen-j:
…: link to the wiki page with the rest of the vision document

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,38 @@
-# esque - an operational kafka tool.
+# esque - an operational Kafka ecosystem administration tool
 
 [![pypi Version](https://img.shields.io/pypi/v/esque.svg)](https://pypi.org/project/esque/) [![Python Versions](https://img.shields.io/pypi/pyversions/esque.svg)](https://pypi.org/project/esque/) [![Build Status](https://travis-ci.org/real-digital/esque.svg?branch=master)](https://travis-ci.org/real-digital/esque) [![Coverage Status](https://coveralls.io/repos/github/real-digital/esque/badge.svg)](https://coveralls.io/github/real-digital/esque?branch=add-coverage) [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
 
-In the Kafka world nothing is easy, but `esque` (pronounced esk) is an attempt at it.
+In the Kafka world nothing is easy, but `esque` (pronounced *esk*) is an attempt at it.
+
+Esque is a user-centric command line interface for Kafka administration. 
 
 ## Why should you care?
 
 Some stuff is hard, and that is okay, listing your kafka topics shouldn't be.
 
 While adopting kafka at real.digital we noticed the immense entry barrier it poses to newcomers. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,27 +1,38 @@
-# esque - an operational kafka tool.
+# esque - an operational Kafka ecosystem administration tool
```
hfjn: Actually I think I like the old line heading better. :)

README.md
```diff
@@ -1,27 +1,38 @@
...
-## Main Features
-
-* List Resources (Topics, Consumergroups, Brokers)
-* Get detailed Overviews of Resources (Topics, Consumergroups, Brokers)
-* Create/Delete Topics
-* Context Switch (Easily Switch between pre-defined Clusters)
+We can't recount how often we wrote Slack messages asking for the script to check the 
+status of topics or consumer groups. This is partly (but not only) due to a 
+fragmented and unclear definition of tooling and APIs for kafka. 
+In a wide array of administration tools, `esque` distances itself by striving to provide Kafka Ops for Humans, in a usable and natural way. We feel that the goal of Esque embodies one of the Perl design principles: “**keep easy things easy, and make hard things possible**”.
```
hfjn: I think I would prefer to write esque with a lowercase letter. But I'm not completely set.

README.md
```diff
@@ -1,27 +1,38 @@
...
-* Context Switch (Easily Switch between pre-defined Clusters)
+We can't recount how often we wrote Slack messages asking for the script to check the 
+status of topics or consumer groups. This is partly (but not only) due to a 
+fragmented and unclear definition of tooling and APIs for kafka. 
+In a wide array of administration tools, `esque` distances itself by striving to provide Kafka Ops for Humans, in a usable and natural way. We feel that the goal of Esque embodies one of the Perl design principles: “**keep easy things easy, and make hard things possible**”. 
+
+## Principles
+
+Our driving objectives can be summed up through these four principles:
+* comes with batteries included,
```
swenzel: What does this actually mean? That you can run a cluster with esque? That you don't have to install additional dependencies?

README.md
swenzel: For me `Esque` looks a little strange too, but mainly because we're so used to the lowercase variant. Maybe we can agree on putting it lowercase in backticks?

README.md
```diff
@@ -1,27 +1,38 @@
...
+* feature rich,
+* robust,
+* insightful,
+* made by engineers for engineers
+
+## Basic Features
+
+* Support for any type of Kafka deployment >1.2,
```
swenzel: I feel like trailing commas in a bullet list are not necessary, same for the list above.

README.md
ognjen-j: The meaning definitely includes your second question. I wanted it to include the first one as well, but others thought such functionality was beyond the purpose of a Kafka administration tool (as it would involve too much system administration).

README.md
ognjen-j: Fixed them. As well as all uppercase "Esque"s

README.md
ognjen-j: These are basically typos. Fixed.

swenzel (approved):

PR #72, opened 2019-09-23
Implement protocol

### LINKED ISSUE #66: Implement own version of binary kafka protocol
None of the current kafka protocol implementations is feature complete. To minimize dependencies on other libraries we could implement our own protocol. This should also enable #51

swenzel:
Consider this a draft, there is still much work to be done here.
Also I doubt that all API endpoints work out of the box. We should add tests for them.

closes #66


hfjn:
What??? You kidding me right? I didn’t think you were that far. 😱

PR #76, opened 2019-09-24
added method for local topics

### LINKED ISSUE #74: get topics takes forever
Currently `esque get topics` takes a huge amount of time because `list_topics` actually enriches all topic objects. Here it should be enough to just retrieve the topic_names in the get use case.

annasweb:
closes #74

### DIFF
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -56,7 +56,7 @@ class TopicController:
         if sort:
             topic_names = sorted(topic_names)
 
-        topics = list(map(self.get_cluster_topic, topic_names))
+        topics = list(map(lambda topic_name: self.get_local_topic(topic_name, False), topic_names))
         return topics
 
     @raise_for_kafka_exception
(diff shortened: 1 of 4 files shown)


esque/controller/topic_controller.py
```diff
@@ -56,7 +56,7 @@ def list_topics(self, *, search_string: str = None, sort: bool = True, hide_inte
         if sort:
             topic_names = sorted(topic_names)
 
-        topics = list(map(self.get_cluster_topic, topic_names))
+        topics = list(map(lambda topic_name: self.get_local_topic(topic_name, False), topic_names))
```
hfjn: We misunderstood each other I think. I would have introduce a flag here. So the `list_topic` can be used to either collect all topic_names or all topic objects. and just use `get_cluster_topic` or the `get_local_topic`based on the use case. Than you would just call the method with the flag set in the get topics command. :)

esque/controller/topic_controller.py
```diff
@@ -93,6 +93,11 @@ def get_cluster_topic(self, topic_name: str) -> Topic:
         """Convenience function getting an existing topic based on topic_name"""
         return self.update_from_cluster(Topic(topic_name))
 
+    def get_local_topic(self, topic_name: str, get_from_cluster: bool = True) -> Topic:
+        if get_from_cluster:
```
hfjn: If you change the function like I mentioned above you don't need the flag here. :)

hfjn (changes requested):

hfjn (approved):

PR #79, opened 2019-09-26
Improve exception handling

### LINKED ISSUE #75: Improve Exception Handling
- Never expose python exceptions to user by
- catching all possible exceptions and return a user error message

Bibob7:
closes #75

### DIFF
--- a/esque/cli/commands.py
+++ b/esque/cli/commands.py
@@ -1,5 +1,4 @@
 import pathlib
-import sys
 import time
 from pathlib import Path
 from shutil import copyfile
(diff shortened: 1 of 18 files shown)


esque/cli/commands.py
```diff
@@ -265,20 +261,18 @@ def describe_broker(state, broker_id):
 
 @describe.command("consumergroup")
 @click.argument("consumer-id", required=False)
-@click.option("-v", "--verbose", help="More detailed information.", default=False, is_flag=True)
+@click.option("--raw-output", help="Get raw output", default=False, is_flag=True)
```
```suggestion
@click.option("--all-partitions", help="List status for all topic partitions instead of just summarizing each topic.", default=False, is_flag=True)
```

swenzel (changes requested):
Could you also rename `raise_for_kafka_exception` to `translate_third_party_exceptions`

swenzel (approved):

swenzel:
The only thing that can be improved is the `--verbose` option during testing... there might be a better way to automatically set this to true or maybe monkeypatch the exception handler during tests. But we can solve this later. I want the exceptions now :D

swenzel:
closes #75

PR #82, opened 2019-10-08
Get last message time

### LINKED ISSUE #48: Get last message time
In order to find out how outdated a topic is, it can be useful to output the timestamp of the last message in that topic.

annasweb:
Closes #48


esque/cli/commands.py
```diff
@@ -232,7 +233,22 @@ def describe_topic(state, topic_name):
         topic = state.cluster.topic_controller.get_cluster_topic(topic_name)
         config = {"Config": topic.config}
 
+        high_watermark = 0
```
hfjn: Maybe move the `max_offset` functionality into it's own method and put it in the appropriate place in topic.py. Makes it also easier to test this.

esque/cli/commands.py
```diff
@@ -232,7 +233,22 @@ def describe_topic(state, topic_name):
...
+        high_watermark_partition_id = 0
+
+        for partition_id in topic.offsets:
+            if topic.offsets[partition_id].high > high_watermark:
+                high_watermark = topic.offsets[partition_id].high
+                high_watermark_partition_id = partition_id
+
+        consumer = MessageConsumer(
```
hfjn: Same holds for the last_message functionality. Maybe something for `topic_controller`?

hfjn (changes requested):
Nice work. 👍 Two minor changes. Please also add tests where possible.

esque/controller/topic_controller.py
```diff
@@ -120,8 +121,24 @@ def update_from_cluster(self, topic: Topic):
...
 
+        message = self._get_last_message(topic)
+        topic.last_message_timestamp = message.timestamp()[1]
+
         return topic
 
+    def _get_last_message(self, local_topic: Topic) -> Optional[Tuple[str, int]]:
+        high_watermark = local_topic.max_offset()["offset"] - 1
+        consumer = MessageConsumer(
+            PING_GROUP_ID,
```
annasweb: I'm not sure about the usage of this group, if you know a better one please tell me

esque/resources/topic.py
```diff
@@ -133,6 +133,7 @@ def __init__(
         num_partitions: int = None,
         replication_factor: int = None,
         config: Dict[str, str] = None,
+        last_message_timestamp: int = None,
```
swenzel:
```suggestion
```

esque/controller/topic_controller.py
```diff
@@ -120,8 +121,24 @@ def update_from_cluster(self, topic: Topic):
 
         topic.is_only_local = False
 
+        message = self._get_last_message(topic)
+        topic.last_message_timestamp = message.timestamp()[1]
+
         return topic
 
+    def _get_last_message(self, local_topic: Topic) -> Optional[Tuple[str, int]]:
```
swenzel: The message with the highest offset isn't necessarily also the latest message. You'll have to go through all partitions and look at the last message to find out which one is the latest in time.

esque/controller/topic_controller.py
swenzel: For now that's okay I think. Maybe we'll have some anonymous consumer at some point, since for the ping we don't need a full blown consumer group either but until then we can reuse the same group id.

swenzel (changes requested):

esque/controller/topic_controller.py
annasweb: Ah okay that's good to know. It makes the code even "smaller" because I won't need the `max_offset` method in `topic.py`

hfjn: Oh that's right. Fuck up on my side. Sorry. 😇

esque/controller/topic_controller.py
```diff
@@ -122,24 +124,29 @@ def update_from_cluster(self, topic: Topic):
...
-        consumer = MessageConsumer(
-            PING_GROUP_ID,
-            local_topic.name,
-            True,
-            starting_offset=high_watermark,
-            partition=local_topic.max_offset()["partition_id"],
-        )
-        message = consumer.consume()
+    @translate_third_party_exceptions
+    def _get_last_message(self, local_topic: Topic) -> Message:
```
swenzel: How about adding `last_message_timestamp` to the Partition object. Then this could just be `max(partition.last_message_timestamp for partition in local_topic.partition_data)`.
If you then add it to `Partition.as_dict()` it will also work nicer with the standardized output we're about to introduce.

annasweb: No problem, I just thought that the issue was to show the last message for each topic, but adding it to the partition makes the code even easier, again ;)

hfjn (approved):
LGTM. 👍

hfjn:
But please fix the tests before merging. :)

swenzel (approved):

PR #83, opened 2019-10-09
73 validity checks for edit

b-boehne:
comparison with `improve-exception-handling` to not see Kevins changes which i merged in my branch to use his error handling

will of course merge to `master` after approval


esque/resources/topic.py
```diff
@@ -6,12 +6,68 @@
...
+from esque.errors import TopicConfigNotValidException
 
 TopicDict = Dict[str, Union[int, str, Dict[str, str]]]
 PartitionInfo = Dict[int, OffsetPartitionResponse]
 
 Watermark = namedtuple("Watermark", ["high", "low"])
 
+# kafka 2.3 - value of dict is either enum or type
+allowed_configs = {
+    "cleanup.policy": ["compact", "delete"],
```
swenzel: Other allowed values are `""`,`"compact,delete"` and `"delete,compact"`.

swenzel (changes requested):
I think for now this is good, but in future we might want to use a more powerful validator. Like https://github.com/23andMe/Yamale for example.

b-boehne:
> I think for now this is good, but in future we might want to use a more powerful validator. Like https://github.com/23andMe/Yamale for example.

damn I reinvented the wheel again, this is perfect for our use case
are you sure we don't want to use yamale from the get-go? i'd be willing to do it

swenzel:
Go for it :)

b-boehne:
> Go for it :)

The problem I currently have is that all configs are saved as strings. I can't really validate all of them safely with typecasts or something without using RegEx for all of those strings, but it's probably going to be slow.
What do you think? Less strict or more slow? :sweat_smile:

swenzel:
You could write https://github.com/23andMe/Yamale#custom-validators like `sint` that make sure the value is a string and then parse them to int and use the default int validator. Or extend the `int` validator so it takes the argument `as_string=True` telling it to assert the value is a string and then treat it as an integer.

esque/errors.py
```diff
@@ -93,6 +93,23 @@ class TopicDoesNotExistException(KafkaException):
...
 
+class TopicConfigNotValidException(ExceptionWithMessage):
+    def __init__(self, message):
+        self.message = message
+
+    def describe(self) -> str:
+        return self.message
+
+
+class YamaleValidationException(TopicConfigNotValidException):
```
swenzel: I'd put the inheritance the other way around since we might use Yamale for other validations as well.

esque/validation/validators.py
```diff
@@ -0,0 +1,71 @@
...
+    validators[StringFloat.tag] = StringFloat
+    validators[StringDictionary.tag] = StringDictionary
+
+    return validators
+
+
+def catch_value_errors(validate):
+    """method decorator to catch ValueErrors for casts and return an error"""
+
+    def wrapper(self, value):
```
swenzel, line 21:
```suggestion
    @wraps(validate)
    def wrapper(self, value):
```
`wraps` is from the `functools` module.

esque/validation/validators.py
```diff
@@ -0,0 +1,71 @@
...
+
+class StringFloat(Number):
+    tag = "s_float"
+
+    @catch_value_errors
+    def validate(self, value):
+        return super().validate(float(value))
+
+
+class StringDictionary(Validator):
```
swenzel: That's actually a special case of a "dictionary". Therefore I'd rather choose a different name.
Or use regex for it. Btw. `'*'` is also a valid value for the `throttled.replicas` fields.
The regex might be: https://regex101.com/r/Phaz0a/1

b-boehne: Even though I'm a big fan of regex, I'd always go for more maintainable solutions like I think I did.
Renaming and adding * makes sense though.

swenzel: Well, you _do_ use regex in your validator :stuck_out_tongue:. Also `KeyValuePair` still suggests something more general than it actually is. Only pairs of numbers are allowed. And a '*'. And an empty string. Pretty specific if you ask me :D  
How about calling it `ReplicaList`?

Btw, did you know you can do:
```yaml
config:
  follower.replication.throttled.replicas: include('replica_list')
  leader.replication.throttled.replicas: include('replica_list')
---
replica_list: regex('^(?:\*|\d+:\d+(?:,\d+:\d+)*|)$')
```
Not suggesting you have to, I'm fine with using your validator as long as it's name is more descriptive. :)

swenzel (approved):

hfjn (approved):

PR #91, opened 2019-10-18
84 split transfer into consume produce

ognjen-j:


esque/clients/consumer.py
```diff
@@ -17,33 +18,56 @@
...
                 "enable.auto.commit": True,
-                "enable.partition.eof": False,
+                "enable.partition.eof": True,
                 # We need this to start at the last committed offset instead of the
                 # latest when subscribing for the first time
                 "default.topic.config": {"auto.offset.reset": offset_reset},
             }
         )
+
+    @abstractmethod
```
hfjn, line 59: If it is abstract. Should it have any implementation?

esque/clients/producer.py
```diff
@@ -2,69 +2,138 @@
...
+        for single_message_line in sys.stdin:
+            message_valid = False
+            try:
+                message = deserialize_message(single_message_line)
+                message_valid = True
+            except JSONDecodeError:
+                if self._ignore_errors:
+                    message = KafkaMessage(key=None, value=single_message_line, partition=0)
+                    message_valid = True
+            finally:
```
hfjn: You don't need the finally. The if statement is enough. :)

esque/clients/producer.py
```diff
@@ -2,69 +2,138 @@
...
+
+    @translate_third_party_exceptions
+    def produce(self) -> int:
+        total_number_of_produced_messages = 0
+        # accept lines from stdin until EOF
+        for single_message_line in sys.stdin:
+            message_valid = False
+            try:
+                message = deserialize_message(single_message_line)
+                message_valid = True
```
hfjn: super nitpicky: If you just instantiate message with None each loop you can check for `if message` and save a whole of two lines. 😝

esque/clients/consumer.py
```diff
@@ -53,47 +77,105 @@ def _subscribe(self, topic: str) -> None:
...
+        return counter
+
+    def create_internal_consumer(self):
+        super()._setup_config()
+        super().create_internal_consumer()
+        if self._topic_name is not None:
+            self._subscribe(self._topic_name)
+
+    def output_consumed(self, message: Message, file_writer: FileWriter = None):
+        print(serialize_message(message))
```
hfjn: `click.echo` instead of print?

esque/clients/producer.py
```diff
@@ -2,69 +2,138 @@
...
     @translate_third_party_exceptions
-    def produce(self, topic_name: str) -> int:
+    def produce(self) -> int:
         start = pendulum.now()
-        self._producer.produce(topic=topic_name, key=str(0), value=str(pendulum.now().timestamp()))
+        self.produce_message(
+            topic_name=self._topic_name,
+            message=KafkaMessage(key=str(0), value=str(pendulum.now().timestamp()), partition=0),
+        )
         while True:
```
hfjn: Why not overwrite the `flush_all` here?

esque/messages/message.py
```diff
@@ -51,34 +55,48 @@ def __init__(self, directory: pathlib.Path):
...
+        "partition": decoded_message.partition,
+    }
+    return json.dumps(serializable_message)
+
+
+def deserialize_message(message_line: str) -> KafkaMessage:
+    json_record = json.loads(message_line)
+    key = None if "key" not in json_record or not json_record["key"] else json_record["key"]
+    value = json_record["value"]
+    partition = json_record["partition"] if "partition" in json_record else 0
```
hfjn:
```suggestion
    partition = json_record.get("partition", 0)
```

tests/integration/test_clients.py
```diff
@@ -108,6 +119,119 @@ def test_avro_consume_and_produce(
     assert produced_messages == consumed_messages
 
 
+@pytest.mark.integration
+def test_plain_text_message_ordering(
+    producer: ConfluenceProducer, topic_multiple_partitions: str, tmpdir_factory, cli_runner: CliRunner
+):
+    working_dir = tmpdir_factory.mktemp("working_directory")
+    ordered_messages = [
```
hfjn: Please move `ordered_messages` to a fixture or constant.

tests/integration/test_clients.py
```diff
@@ -108,6 +119,119 @@ def test_avro_consume_and_produce(
...
+        KafkaMessage(key="h", value="v3", partition=2),
+        KafkaMessage(key="g", value="v4", partition=3),
+        KafkaMessage(key="f", value="v5", partition=2),
+        KafkaMessage(key="e", value="v6", partition=1),
+        KafkaMessage(key="d", value="v7", partition=0),
+        KafkaMessage(key="c", value="v8", partition=2),
+        KafkaMessage(key="b", value="v9", partition=3),
+        KafkaMessage(key="a", value="v10", partition=1),
+    ]
+    for message in ordered_messages:
```
hfjn: Actually you could just move the whole test setup to a fixture

tests/integration/test_clients.py
```diff
@@ -108,6 +119,119 @@ def test_avro_consume_and_produce(
...
+    )
+    # Check assertions:
+    consumed_messages = get_consumed_messages(working_dir, False, sort=False)
+    assert consumed_messages[0].key == "i"
+    assert consumed_messages[1].key == "e"
+    assert consumed_messages[2].key == "a"
+
+
+@pytest.mark.integration
+def test_plain_text_message_cli_pipe(
```
hfjn: please move the cli test to `integration/commands/test_consume.py`

hfjn (changes requested):

hfjn (approved):

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
                 # We need this to start at the last committed offset instead of the
                 # latest when subscribing for the first time
                 "default.topic.config": {"auto.offset.reset": offset_reset},
             }
         )
-        self._consumer = confluent_kafka.Consumer(self._config)
+
+    @abstractmethod
+    def create_internal_consumer(self):
+        pass
```
swenzel:
```suggestion
        raise NotImplementedError()
```
This is the python way of saying: "Please implement this method in your subclass"

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
+
+    @translate_third_party_exceptions
+    def assign_specific_partitions(self, topic_name: str, partitions: list = None, offset: int = 0):
         self._topic_name = topic_name
+        if partitions is not None:
+            topic_partitions = [
+                TopicPartition(self._topic_name, partition=partition, offset=offset) for partition in partitions
+            ]
+        else:
+            topic_partitions = [TopicPartition(self._topic_name, 0, 0)]
```
swenzel:
```suggestion
            topic_partitions = [TopicPartition(self._topic_name, partition=0, offset=offset)]
```

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+        """
+        Outputs the message to a destination determined by the implementation of the inheriting class.
+        :param message: Message to output
+        :return: This method returns no values
+        """
+        writer = self.writers.get(message.partition(), self.writers[-1])
+        writer.write_message(message)
+
+    def close_all_writers(self):
+        for _, w in self.writers.items():
```
swenzel:
```suggestion
        for w in self.writers.values():
```

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+    def consume_single_message(self, timeout=30) -> Message:
         message = self._consumer.poll(timeout=timeout)
         raise_for_message(message)
         return message
 
+    @translate_third_party_exceptions
+    def consume_single_acceptable_message(self, timeout=30) -> Message:
+        message_acceptable = False
+        while not message_acceptable:
+            message = self._consumer.poll(timeout=timeout)
```
swenzel: In order to keep the method timeout correct, you should subtract the elapsed time from the timeout you pass to `poll` here and raise a timeout exception if the remaining time is 0 or less. Otherwise you can wait multiple times for `timeout` seconds.  
Also, why not use `self._consume_single_message`?

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
-        if self.rule_tree is not None:
-            return self.rule_tree.evaluate(message)
+        if self._rule_tree is not None:
+            return self._rule_tree.evaluate(message)
         else:
             return True
 
 
 class PingConsumer(AbstractConsumer):
     def __init__(self, group_id: str, topic_name: str, last: bool):
```
swenzel: Signature and code of the superclass are compatible with this one, so you don't need to define a constructor here.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
 
-        self._config.update({"default.topic.config": {"auto.offset.reset": offset_reset}})
-        self._consumer = confluent_kafka.Consumer(self._config)
-        self._subscribe(topic_name)
+
+class PlaintextConsumer(AbstractConsumer):
+    def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, last: bool, match: str = None):
+        super().__init__(group_id, topic_name, last, match)
+        self.working_dir = working_dir
+        super()._setup_config()
```
swenzel:
```suggestion
```
Parent constructor will handle this.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+    def create_internal_consumer(self):
+        self._setup_config()
+        self._consumer = confluent_kafka.Consumer(self._config)
+        self.assign_specific_partitions(self._topic_name, partitions=[0], offset=0)
 
-class FileConsumer(AbstractConsumer):
-    def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, last: bool):
-        super().__init__(group_id, topic_name, last)
-        self.working_dir = working_dir
+    def _setup_config(self):
```
swenzel, line 133: You can use the parent implementation and override `enable.partition.eof`.
```
def _setup_config(self):
    super()._setup_config()
    self._config['enable.partition.eof'] = False
```
However, I'm not sure if we actually need to disable partition eof for the ping consumer. So you could just as well remove this implementation entirely.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
                 counter += 1
+                self.output_consumed(message)
 
+        self.close_all_writers()
         return counter
 
-    def get_file_writer(self, partition: int) -> FileWriter:
-        return PlainTextFileWriter((self.working_dir / f"partition_{partition}"))
+    def create_internal_consumer(self):
+        super()._setup_config()
```
swenzel:
```suggestion
```
Already handled by constructor.

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
+        self._config = Config().create_confluent_config()
+        if self._match is not None:
+            self._rule_tree = RuleTree(match)
+        else:
+            self._rule_tree = None
+        self._topic_name = topic_name
+        self._use_single_directory_for_messages = False
+        self.writers: Dict[int, GenericWriter] = {}
+
+    @translate_third_party_exceptions
```
swenzel:
```suggestion
```
No third party exception to be expected here.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
-        super().__init__(group_id, topic_name, working_dir, last)
+class AvroFileConsumer(PlaintextConsumer):
+    def __init__(self, group_id: str, topic_name: str, working_dir: pathlib.Path, last: bool, match: str = None):
+        super().__init__(group_id, topic_name, working_dir, last, match)
         self.schema_registry_client = SchemaRegistryClient(Config().schema_registry)
-        self._subscribe(topic_name)
 
-    def get_file_writer(self, partition: int) -> FileWriter:
-        return AvroFileWriter((self.working_dir / f"partition_{partition}"), self.schema_registry_client)
+    def create_internal_consumer(self):
```
swenzel: No need to implement this, `PlaintextConsumer.create_internal_consumer` already does this.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+        last: bool,
+        avro: bool,
+        merge_partitions_to_single_file: bool = False,
+        match: str = None,
+    ):
+        if avro:
+            consumer = AvroFileConsumer(
+                group_id=group_id, topic_name=topic_name, working_dir=working_dir, last=last, match=match
+            )
+            consumer.writers[-1] = (
```
swenzel: This looks like some code that belongs into a constructor. The class itself _relies_ on `writers[-1]` so it should be its own responsibility to make sure it exists.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+                StdOutAvroWriter(schema_registry_client=consumer.schema_registry_client)
+                if working_dir is None
+                else AvroFileWriter(consumer.working_dir / "partition_any", consumer.schema_registry_client)
+            )
+        else:
+            consumer = PlaintextConsumer(
+                group_id=group_id, topic_name=topic_name, working_dir=working_dir, last=last, match=match
+            )
+            consumer.writers[-1] = (
```
swenzel: same here

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+                StdOutWriter() if working_dir is None else PlainTextFileWriter(consumer.working_dir / "partition_any")
+            )
+        consumer._use_single_directory_for_messages = merge_partitions_to_single_file
+        if merge_partitions_to_single_file and consumer.working_dir is not None:
+            consumer.writers[-1].init_destination_directory()
+
+        consumer.create_internal_consumer()
```
swenzel:
```suggestion
```
No need to call this here if you follow my suggestion for calling it in the constructor.

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
+        self._producer = None
+        if self._match is not None:
+            self._rule_tree = RuleTree(match)
+        else:
+            self._rule_tree = None
+
+    @translate_third_party_exceptions
+    @abstractmethod
+    def produce(self) -> int:
+        pass
```
swenzel:
```suggestion
        raise NotImplementedError()
```

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
             }
         )
 
+    @translate_third_party_exceptions
+    def flush_all(self, message_prefix: str = None):
+        while True:
+            left_messages = self._producer.flush(1)
+            if left_messages == 0:
+                break
+            click.echo((message_prefix or "") + f"Still {left_messages} messages left, flushing...")
```
swenzel: This could interfere with our formatted output, better use the `logging` module here. I can show you later how to use it.

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
+
     @translate_third_party_exceptions
     @abstractmethod
```
swenzel: This method is not abstract, it has an implementation.
```suggestion
```

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
-                break
-            click.echo(f"{delta_t(start)} | Still {left_messages} messages left, flushing...")
+        self.produce_message(
+            topic_name=self._topic_name,
+            message=KafkaMessage(key=str(0), value=str(pendulum.now().timestamp()), partition=0),
+        )
+        self.flush_all(message_prefix=f"{delta_t(start)} | ")
         return 1
 
+    def create_internal_producer(self):
```
swenzel: You already got that covered, no need for implementation.

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
+                message = deserialize_message(single_message_line.replace("\n", ""))
+            except (JSONDecodeError, KeyError):
+                if self._ignore_errors:
+                    message = KafkaMessage(key=None, value=single_message_line, partition=0)
+            if message:
+                self.produce_message(topic_name=self._topic_name, message=message)
+                total_number_of_produced_messages += 1
+        return total_number_of_produced_messages
+
+    def create_internal_producer(self):
```
swenzel: Already got that covered as well, no need for implementation.

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
-                break
-            click.echo(f"Still {left_messages} messages left, flushing...")
-
     def get_file_reader(self, directory: pathlib.Path) -> FileReader:
         return PlainTextFileReader(directory)
 
-    @translate_third_party_exceptions
-    def produce_message(self, topic_name: str, message: KafkaMessage):
-        self._producer.produce(topic=topic_name, key=message.key, value=message.value, partition=message.partition)
+    def create_internal_producer(self):
```
swenzel: again, no need for implementation

esque/clients/producer.py
```diff
@@ -1,110 +1,194 @@
...
 class AvroFileProducer(FileProducer):
-    def __init__(self, working_dir: pathlib.Path):
-        super().__init__(working_dir)
-        self._config.update({"schema.registry.url": Config().schema_registry})
+    def __init__(self, topic_name: str, working_dir: pathlib.Path, match: str = None):
+        super().__init__(topic_name=topic_name, working_dir=working_dir, match=match)
+
+    @translate_third_party_exceptions
+    def create_internal_producer(self):
+        self._setup_config()
```
swenzel:
```suggestion
```
That's already called during `__init__`

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
 from esque.ruleparser.ruleengine import RuleTree
 
 
 class AbstractConsumer(ABC):
     def __init__(self, group_id: str, topic_name: str, last: bool, match: str = None):
+        self._match = match
+        self._last = last
+        self._group_id = group_id
+        self._consumer = None
+        self._config = Config().create_confluent_config()
```
swenzel:
```suggestion
        self._config = {}
```
Better to do this in `_setup_config()`

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
+        if self._match is not None:
+            self._rule_tree = RuleTree(match)
+        else:
+            self._rule_tree = None
+        self._topic_name = topic_name
+        self._use_single_directory_for_messages = False
+        self.writers: Dict[int, GenericWriter] = {}
```
swenzel, line 36:
```suggestion
        self.writers: Dict[int, GenericWriter] = {}
        self._setup_config()
        self.create_internal_consumer()
```
Better to do this automatically here, so you don't have to repeat yourself within `create_internal_consumer`, also call `create_internal_consumer` directly, since you have to do it anyway.

esque/clients/consumer.py
```diff
@@ -1,49 +1,75 @@
...
+
+    @translate_third_party_exceptions
+    def _setup_config(self):
```
swenzel, line 40:
```suggestion
    def _setup_config(self):
        self._config.update(Config().create_confluent_config())
```

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
     def consume(self, **kwargs) -> int:
         pass
```
swenzel:
```suggestion
        raise NotImplementedError()
```

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
         msg_sent_at = pendulum.from_timestamp(float(message.value()))
         delta_sent = pendulum.now() - msg_sent_at
         return message.key(), delta_sent.microseconds / 1000
 
-    @translate_third_party_exceptions
-    def _assign_exact_partitions(self, topic: str) -> None:
-        self._consumer.assign([TopicPartition(topic=topic, partition=0, offset=0)])
-
+    def create_internal_consumer(self):
+        self._setup_config()
```
swenzel:
```suggestion
```
Constructor will handle this.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+            consumer.writers[-1].init_destination_directory()
+
+        consumer.create_internal_consumer()
+        return consumer
+
+    def create_ping_consumer(self, group_id: str, topic_name: str):
+        consumer = PingConsumer(group_id, topic_name, last=False)
+        consumer._uses_file_output = False
+        consumer.writers[0] = StdOutWriter()
+        consumer.create_internal_consumer()
```
swenzel:
```suggestion
```
See above

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+        consumer._use_single_directory_for_messages = merge_partitions_to_single_file
+        if merge_partitions_to_single_file and consumer.working_dir is not None:
+            consumer.writers[-1].init_destination_directory()
+
+        consumer.create_internal_consumer()
+        return consumer
+
+    def create_ping_consumer(self, group_id: str, topic_name: str):
+        consumer = PingConsumer(group_id, topic_name, last=False)
+        consumer._uses_file_output = False
```
swenzel:
```suggestion
```
`PingConsumer` doesn't have that attribute.

esque/clients/consumer.py
```diff
@@ -53,81 +79,181 @@ def _subscribe(self, topic: str) -> None:
...
+
     @translate_third_party_exceptions
-    def _consume_single_message(self, timeout=30) -> Message:
+    def consume_single_message(self, timeout=30) -> Message:
         message = self._consumer.poll(timeout=timeout)
         raise_for_message(message)
         return message
 
+    @translate_third_party_exceptions
+    def consume_single_acceptable_message(self, timeout=30) -> Message:
```
swenzel:
```suggestion
    def consume_single_acceptable_message(self, timeout=30) -> Optional[Message]:
```

esque/cli/commands.py
```diff
@@ -328,59 +322,190 @@ def get_topics(state: State, prefix: str, output_format: str):
...
+@click.option("-n", "--numbers", help="Number of messages", type=click.INT, default=sys.maxsize, required=False)
+@click.option("-m", "--match", help="Message filtering expression", type=click.STRING, required=False)
 @click.option("--last/--first", help="Start consuming from the earliest or latest offset in the topic.", default=False)
 @click.option("-a", "--avro", help="Set this flag if the topic contains avro data", default=False, is_flag=True)
 @click.option(
-    "-k",
-    "--keep",
-    "keep_file",
-    help="Set this flag if the file with consumed messages should be kept.",
+    "-o",
```
```suggestion
    "-p",
```
Other commands use `-o`  for `--output-format` which could lead to confusion.

esque/cli/commands.py
```diff
@@ -328,59 +322,190 @@ def get_topics(state: State, prefix: str, output_format: str):
...
+    messages_by_partition = {}
+    partitions_by_timestamp = {}
+    total_number_of_messages = 0
+    no_more_messages = False
+    # get at least one message from each partition, or exclude those that don't have any messages
+    for partition_counter in range(0, len(consumers)):
+        try:
+            message = consumers[partition_counter].consume_single_acceptable_message(timeout=10)
+            decoded_message = decode_message(message)
+        except (MessageEmptyException, EndOfPartitionReachedException):
```
swenzel: `MessageEmtpyException` could also just indicate a timeout. That doesn't mean that there are no messages, that's why we have `EndOfPartitionReachedException` which is the only way to reliably detect if you've reached the end of a partition.

esque/cli/commands.py
```diff
@@ -328,59 +322,190 @@ def get_topics(state: State, prefix: str, output_format: str):
...
+                no_more_messages = True
+        else:
+            partitions_by_timestamp[decoded_message.timestamp] = decoded_message.partition
+            if decoded_message.partition not in messages_by_partition:
+                messages_by_partition[decoded_message.partition] = []
+            messages_by_partition[decoded_message.partition].append(message)
+
+    # in each iteration, take the earliest message from the map, output it and replace it with a new one (if available)
+    # if not, remove the consumer and move to the next one
+    while total_number_of_messages < numbers and not no_more_messages:
```
swenzel: It's better to avoid negative boolean variable names. `while messages_left` is more readable than `while not no_more_messages`

esque/cli/commands.py
```diff
@@ -328,59 +322,190 @@ def get_topics(state: State, prefix: str, output_format: str):
...
+            click.echo(
+                "Found only "
+                + bold(str(total_number_of_consumed_messages))
+                + " messages in topic, out of "
+                + blue_bold(str(numbers))
+                + " required."
+            )
+
+
+def _consume_to_file_ordered(
```
swenzel: Why is this code not part of the consumer? Or at least the consumer module.

esque/cli/commands.py
```diff
@@ -391,51 +516,86 @@ def transfer(
...
     help="Sets the directory that contains Kafka messages",
     type=click.STRING,
-    required=True,
+    required=False,
 )
+@click.option("-t", "--to", "to_context", help="Destination Context", type=click.STRING, required=False)
+@click.option("-m", "--match", help="Message filtering expression", type=click.STRING, required=False)
 @click.option("-a", "--avro", help="Set this flag if the topic contains avro data", default=False, is_flag=True)
+@click.option(
+    "-i",
```
swenzel: Either remove `-i` or add some equivalent to `consume` in order to stay consistent.

esque/clients/producer.py
```diff
@@ -12,99 +14,175 @@
...
         self.internal_queue_length_limit = self.queue_length / 0.5
         self._config = Config().create_confluent_config()
+        self._setup_config()
+        self._topic_name = topic_name
+        self._match = match
+        self._producer = None
+        if self._match is not None:
+            self._rule_tree = RuleTree(match)
+        else:
+            self._rule_tree = None
```
swenzel, line 34:
```suggestion
            self._rule_tree = None
            self.create_internal_producer()
```

esque/clients/producer.py
```diff
@@ -12,99 +14,175 @@
...
 
 
-class PingProducer(Producer):
-    def __init__(self):
-        super().__init__()
-        self._producer = confluent_kafka.Producer(self._config)
+class PingProducer(AbstractProducer):
+    def __init__(self, topic_name: str):
+        super().__init__(topic_name=topic_name)
+        self.create_internal_producer()
```
swenzel:
```suggestion
```
Super constructor will do this if you apply my suggestion.

esque/clients/producer.py
```diff
@@ -12,99 +14,175 @@
...
+    def create_internal_producer(self):
+        super()._setup_config()
+        super().create_internal_producer()
 
 
 class AvroFileProducer(FileProducer):
-    def __init__(self, working_dir: pathlib.Path):
-        super().__init__(working_dir)
-        self._config.update({"schema.registry.url": Config().schema_registry})
+    def __init__(self, topic_name: str, working_dir: pathlib.Path, match: str = None):
```
swenzel: No need for this constructor, it's just passing everything to the parent's constructor so you don't actually need it.

esque/clients/producer.py
```diff
@@ -12,99 +14,175 @@
...
+        match: str = None,
+        ignore_stdin_errors: bool = False,
+    ):
+        if working_dir is None:
+            producer = StdInProducer(topic_name=topic_name, match=match, ignore_errors=ignore_stdin_errors)
+        elif avro:
+            producer = AvroFileProducer(topic_name=topic_name, working_dir=working_dir, match=match)
+        else:
+            producer = FileProducer(topic_name=topic_name, working_dir=working_dir, match=match)
+        producer.create_internal_producer()
```
swenzel:
```suggestion
```
The producer's constructor will call this.

esque/ruleparser/expressionelement.py
```diff
@@ -1,9 +1,10 @@
...
-from enum import Enum
 from collections import namedtuple
+from enum import Enum
+
 import pendulum
-import esque.ruleparser.helpers as h
 
+import esque.ruleparser.helpers as h
 
 GenericOperator = namedtuple(
```
swenzel: If you use `typing.NamedTuple` you can assign types to the fields.

swenzel (changes requested):

esque/cli/commands.py
ognjen-j: Probably best to remove it, since other letters in "stdin" don't really make sense here.

esque/cli/commands.py
ognjen-j: Consumer module fits better, since the methods don't really rely on a single consumer.

esque/cli/commands.py
ognjen-j: As with the consume command, better to leave out the short option completely and then implement a "-" as a substitute for stdin/stdout.

esque/clients/consumer.py
ognjen-j: Less processing is required when reading messages if eof is disabled. This only holds for the PingConsumer.

esque/ruleparser/fieldeval.py
```diff
@@ -28,25 +30,46 @@ def evaluate_field(self, field_name: str):
...
+                    elif isinstance(self.__message, KafkaMessage):
+                        for message_header in self.__message.headers:
+                            if message_header.key == header_name and message_header.value:
+                                return message_header.value
+                            elif message_header.key == header_name:
+                                return ""
                         return ""
-                    else:
-                        return header_value
+                return -1
```
swenzel: You can dedent this by two levels in order to satisfy flake8 :)
```python
if a:
    print('a')
else:
    if b:
        print('b')
    else:
        print('c')
```
is the same as
```python
if a:
    print('a')
elif b:
    print('b')
else:
    print('c')
```

swenzel (approved):

esque/ruleparser/fieldeval.py
ognjen-j: Simplified.

PR #97, opened 2019-11-06
Improve error handling

### LINKED ISSUE #5: Nicer exception handling.
Exceptions like #2 should be caught more nicely. A default python stacktrace is just not nice to look at or understand.

### LINKED ISSUE #86: Refactor errors.py

swenzel:
closes #5 
closes #86


esque/cli/options.py
```diff
@@ -34,10 +37,37 @@ def cluster(self):
             self._cluster = Cluster()
         return self._cluster
 
+    def _get_verbose(self) -> bool:
+        if "ESQUE_VERBOSE" in os.environ:
```
hfjn: We have https://github.com/real-digital/esque/blob/master/esque/cli/environment.py for these. :)

esque/errors.py
```diff
@@ -154,136 +139,5 @@ class TopicConfigNotValidException(YamaleValidationException):
     3: TopicDoesNotExistException,
     36: TopicAlreadyExistsException,
     -191: EndOfPartitionReachedException,
+    38: InvalidReplicationFactorException,
 }
-
-
-class ErrorCode(Enum):
```
hfjn, line 160: Can we keep those in a comment? For future reference.

tests/conftest.py
```diff
@@ -226,7 +232,7 @@ def messages_ordered_different_partition_with_headers() -> Iterable[KafkaMessage
 
 @pytest.fixture()
 def produced_messages_different_partitions(messages_ordered_different_partitions: Iterable[KafkaMessage]):
-    def _produce(topic_name: str, producer: ConfluenceProducer):
+    def _produce(topic_name: str, producer: ConfluentProducer):
```
hfjn, line 234: Hihi. Atlassian at it again.

hfjn (approved):

esque/cli/options.py
swenzel: Yeah... this is just impossible to mock if you have `from esque.cli.environment import ENV_VAR` all over the place. You'll have to find all imports and mock them. Unless you somehow manage to enforce that its just `from esque.cli import environment` everywhere and we only access the vars via the module.
That's why I wanted to avoid it. But I see why we should keep it consistent... I'll change it to use that module and put a comment about mocking the variables there.

esque/errors.py
swenzel: how about I do not use the plain ints in `ERROR_LOOKUP` but the constants from `KafkaError`? Then people can see where it comes from and look there.

tests/conftest.py
swenzel: :smile:

PR #105, opened 2019-11-13
100 improve context help text

garrettthomaskth:


esque/cli/commands.py
```diff
@@ -33,11 +33,15 @@
 from esque.resources.topic import Topic, copy_to_local
 
 
-@click.group(help="esque - an operational kafka tool.", invoke_without_command=True)
+@click.group(invoke_without_command=True)
 @click.option("--recreate-config", is_flag=True, default=False, help="Overwrites the config with the sample config.")
 @version_option(__version__)
 @default_options
 def esque(state: State, recreate_config: bool):
+    """esque - an operational kafka tool.
```
hfjn, line 42: Can we make it possible to show the help when `esque` is called without any options or arguments? Currently nothing is shown.

esque/cli/commands.py
```diff
@@ -382,21 +446,40 @@ def get_consumergroups(state: State, output_format: str):
...
+    type=click.STRING,
+    required=False,
+)
+@click.option(
+    "-n", "--numbers", metavar="<n>", help="Number of messages", type=click.INT, default=sys.maxsize, required=False
+)
+@click.option(
+    "-m",
+    "--match",
+    metavar="<filter_expresion>",
```
hfjn:
```suggestion
    metavar="<filter_expression>",
```

esque/cli/commands.py
```diff
@@ -428,6 +505,24 @@ def consume(
...
     write_to_stdout: bool,
 ):
+    """Consume messages from a topic.
+
+    Read messages from a given topic in a given context. These messages can either be written
+    to files in an automatically generated directory (default behavior), or to STDOUT.
+
+    \b
+    EXAMPLES:
+    #Consume the first 10 messages from TOPIC in the current context and print them to STDOUT in order.
```
hfjn:
```suggestion
    # Consume the first 10 messages from TOPIC in the current context and print them to STDOUT in order.
```

esque/cli/commands.py
```diff
@@ -428,6 +505,24 @@ def consume(
...
+    esque consume --first -n 10 --preserve-order --stdout TOPIC
+
+    \b
+    #Consume <n> messages, starting from the 10th, from TOPIC in the <source_ctx> context and write them to files.
```
hfjn:
```suggestion
    # Consume <n> messages, starting from the 10th, from TOPIC in the <source_ctx> context and write them to files.
```

esque/cli/commands.py
```diff
@@ -428,6 +505,24 @@ def consume(
...
+    esque consume --match "message.offset > 9" -n <n> TOPIC -f <source_ctx>
+
+    \b
+    #Copy source_topic in source_ctx context to destination_topic in destination_ctx
```
hfjn:
```suggestion
    # Copy source_topic in source_ctx context to destination_topic in destination_ctx.
```

hfjn: Should we end help texts on periods or not?

esque/cli/commands.py
```diff
@@ -523,17 +634,35 @@ def produce(
...
     ignore_stdin_errors: bool = False,
 ):
+    """Produce messages to a topic.
+
+       Write messages to a given topic in a given context. These messages can come from either a directory <directory>
+       containing files corresponding to the different partitions or from STDIN.
+
+       \b
+       EXAMPLES:
+       #Write all messages from the files in <directory> to TOPIC in the <destination_ctx> context.
```
hfjn:
```suggestion
       # Write all messages from the files in <directory> to TOPIC in the <destination_ctx> context.
```

esque/cli/commands.py
```diff
@@ -523,17 +634,35 @@ def produce(
...
+       esque produce -d <directory> -t <destination_ctx> TOPIC
+
+       \b
+       #Start environment in terminal to write messages to TOPIC in the <destination_ctx> context.
```
hfjn:
```suggestion
       # Start environment in terminal to write messages to TOPIC in the <destination_ctx> context.
```

esque/cli/commands.py
```diff
@@ -523,17 +634,35 @@ def produce(
...
+       esque produce --stdin -f <destination_ctx> -y TOPIC
+
+       \b
+       #Copy source_topic in source_ctx context to destination_topic in destination_ctx
```
hfjn:
```suggestion
       # Copy source_topic in source_ctx context to destination_topic in destination_ctx.
```

esque/cli/commands.py
```diff
@@ -564,19 +693,24 @@ def produce(
...
 
-@esque.command("ping", help="Tests the connection to the kafka cluster.")
+@esque.command("ping")
 @click.option("-t", "--times", help="Number of pings.", default=10)
 @click.option("-w", "--wait", help="Seconds to wait between pings.", default=1)
 @default_options
 def ping(state: State, times: int, wait: int):
+    """Test the connection to the kafka cluster.
+
+    Ping the kafka cluster by writing messages to and reading messages from the kafka cluster.
```
hfjn:
```suggestion
    Ping the kafka cluster by writing messages to and reading messages from it.
```

esque/cli/commands.py
```diff
@@ -564,19 +693,24 @@ def produce(
...
+    After the specified number of "pings", return the minimum, maximum, and average time.
```
hfjn:
```suggestion
    After the specified number of "pings", return the minimum, maximum, and average time for the round trip.
```

esque/cli/commands.py
```diff
@@ -523,17 +634,35 @@ def produce(
...
+       EXAMPLES:
+       #Write all messages from the files in <directory> to TOPIC in the <destination_ctx> context.
+       esque produce -d <directory> -t <destination_ctx> TOPIC
+
+       \b
+       #Start environment in terminal to write messages to TOPIC in the <destination_ctx> context.
+       esque produce --stdin -f <destination_ctx> -y TOPIC
+
+       \b
+       #Copy source_topic in source_ctx context to destination_topic in destination_ctx.
```
ognjen-j:
```suggestion
       # Copy source_topic in source_ctx context to destination_topic in destination_ctx.
```

esque/cli/commands.py
swenzel: Yes!

esque/cli/commands.py
```diff
@@ -68,13 +72,13 @@ def delete(state: State):
     pass
```
swenzel, line 70:
```suggestion
    ctx = click.get_current_context()
    click.echo(ctx.get_help())
```

esque/cli/commands.py
```diff
@@ -143,26 +159,36 @@ def config_autocomplete(state: State):
...
 
 
-@config.command("migrate", help="Migrate your config to current version")
+@config.command("migrate")
 @default_options
 def config_migrate(state: State):
+    """Migrate esque config to current version.
+
+    If the user's esque config file is from a previous version, migrate it to the current version.
+    A backup of the original config file is created in the with the suffix .bak"""
```
```suggestion
    A backup of the original config file is created with the same name and the extension .bak"""
```

esque/cli/commands.py
```diff
@@ -204,23 +235,37 @@ def edit_topic(state: State, topic_name: str):
...
+    click.echo(click.style(f"Topic with name '{topic_name}' successfully deleted.", fg="green"))
 
 
-@esque.command("apply", help="Apply a configuration")
-@click.option("-f", "--file", help="Config file path", required=True)
+@esque.command("apply")
+@click.option("-f", "--file", metavar="<file>", help="Config file path.", required=True)
 @default_options
 def apply(state: State, file: str):
+    """Apply a configuration.
```
```suggestion
    """Apply a set of topic configurations.
```

esque/cli/commands.py
```diff
@@ -428,6 +505,24 @@ def consume(
...
+    #Consume the first 10 messages from TOPIC in the current context and print them to STDOUT in order.
+    esque consume --first -n 10 --preserve-order --stdout TOPIC
+
+    \b
+    #Consume <n> messages, starting from the 10th, from TOPIC in the <source_ctx> context and write them to files.
+    esque consume --match "message.offset > 9" -n <n> TOPIC -f <source_ctx>
+
+    \b
+    #Copy source_topic in source_ctx context to destination_topic in destination_ctx.
+    esque consume -f source_ctx --stdout source_topic | esque produce -t destination_ctx --stdin destination_topic
```
swenzel: Have we actually tried that? I'm afraid the calls to `config.switch_context` might interefere with each other since they change the actual file and we have a few `Config()` invocations throughout the code which always load the context which was last set in the file. So either pass the config around or make it possible to switch the context without writing the switch back to the file.

esque/cli/commands.py
```diff
@@ -382,39 +446,52 @@ def get_consumergroups(state: State, output_format: str):
...
+    required=False,
+)
+@click.option(
+    "-n", "--numbers", metavar="<n>", help="Number of messages.", type=click.INT, default=sys.maxsize, required=False
+)
+@click.option(
+    "-m",
+    "--match",
+    metavar="<filter_expresion>",
+    help="Message filtering expression.",
```
swenzel, line 610: I think the filter/match expressions might need some documentation, too. Maybe we can introduce another command? Like `esque help filter` ? And then refer the user to that.

esque/cli/commands.py
swenzel: I don't care as long as we stay consistent.

esque/cli/commands.py
garrettthomaskth: It more than doesn't work, it can even break esque :D

swenzel:
My example isn't that cool after all, but click comes to the rescue :smile: 
We can just set `no_args_is_help=True` on the group decorators.  
https://click.palletsprojects.com/en/7.x/api/#click.MultiCommand

esque/cli/commands.py
garrettthomaskth: yesss

esque/cli/commands.py
```diff
@@ -273,7 +273,8 @@ def edit_topic(state: State, topic_name: str):
 @click.option("--offset-by-delta", help="Shift offset by specified value", type=click.INT, required=False)
 @click.option(
     "--offset-to-timestamp",
-    help="Set offset to the value closest to the specified message timestamp in the format YYYY-MM-DDTHH:mm:ss (NOTE: this can be a very expensive operation).",
+    help="Set offset to the value closest to the specified message timestamp in the format YYYY-MM-DDTHH:mm:ss "
```
swenzel: This will not actually put you to the closest value but to the first one at or after that timestamp.
If it were the closest one we'd have to check for the absolute difference in time between two messages surrounding that timestamp.

swenzel: Maybe say something like "Skip over all messages before that timestamp"...

hfjn (approved):

PR #112, opened 2019-11-14
69 describe broker also accepts broker host name

garrettthomaskth:


hfjn (approved):

esque/cli/commands.py
```diff
@@ -417,8 +426,16 @@ def get_watermarks(state: State, topic_name: str, output_format: str):
 @click.argument("broker-id", callback=fallback_to_stdin, autocompletion=list_brokers, required=False)
 @output_format_option
 @default_options
-def describe_broker(state: State, broker_id: str, output_format: str):
-    broker = Broker.from_id(state.cluster, broker_id).describe()
+def describe_broker(state, broker_id, output_format):
```
swenzel: Maybe rename `broker_id` to `broker` and add a help text stating that it actually supports both broker id and host:port combination? Actually we could make the port optional and only fail if there is more than one broker for a given host.

swenzel (approved):

PR #126, opened 2019-11-25
Use default timeout for consume

Bibob7:


esque/resources/topic.py
```diff
@@ -43,7 +42,7 @@ def as_dict(self):
             "partition_isrs": self.partition_isrs,
             "partition_leader": self.partition_leader,
             "partition_replicas": self.partition_replicas,
-            "latest_message_timestamp": self.latest_message_timestamp,
+            "latest_message_timestamp": self.latest_message_timestamp or "Missing due to timeout.",
```
swenzel: `self.last_message_timestamp` will also be `None` when `high_water == low_water`.
I'd just keep it as None. Other logic can get quite complicated.

esque/controller/topic_controller.py
```diff
@@ -180,10 +180,12 @@ def _get_partition_data(
...
             high = int(high_watermarks[partition_id].offset[0])
+            latest_timestamp = None
             if high > low:
-                latest_timestamp = float(consumer.consume(high - 1, partition_id).timestamp()[1]) / 1000
-            else:
-                latest_timestamp = None
+                try:
+                    latest_timestamp = float(consumer.consume(high - 1, partition_id).timestamp()[1]) / 1000
+                except MessageEmptyException:
+                    """Due to timeout latest timestamp is missing."""
```
swenzel: Can you use `log.warning` for that? And include both topic name and partition id?
You'll need to put
```python
import logging
log = logging.getLogger(__name__)
```
At the beginning of the file to get the logger.

poetry.lock
```diff
@@ -1,876 +0,0 @@
-[[package]]
```
swenzel: Why did you remove this file? Put it back please.

swenzel (changes requested):

swenzel (approved):

Issue #127, opened 2019-12-12
Setting consumer group offsets to absolute value of 0 fails


ognjen-j:
Setting consumer offset to an absolute value fails, and the offset is left unchanged. The same operation works when the offset change is denoted by a delta value.

PR #128, opened 2019-12-12
Bugfix.

ognjen-j:
Solves issue #127

### DIFF
--- a/esque/controller/consumergroup_controller.py
+++ b/esque/controller/consumergroup_controller.py
@@ -93,7 +93,7 @@ class ConsumerGroupController:
             self._logger.error("The consumer group {} does not exist.".format(consumer_id))
             return None
         elif consumer_group_state == "Empty":
-            if offset_to_value:
+            if offset_to_value is not None:
                 for plan_element in offset_plans.values():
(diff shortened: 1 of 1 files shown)


esque/controller/consumergroup_controller.py
```diff
@@ -93,15 +93,15 @@ def create_consumer_group_offset_change_plan(
             self._logger.error("The consumer group {} does not exist.".format(consumer_id))
             return None
         elif consumer_group_state == "Empty":
-            if offset_to_value:
+            if offset_to_value is not None:
```
hfjn, line 96: What. That shouldn’t change anything?

hfjn (changes requested):

esque/controller/consumergroup_controller.py
swenzel: Yes it does because `0` is a valid value but doesn't evaluate to `True`. That's why we have to check if it's `None` instead.

swenzel (approved):

esque/controller/consumergroup_controller.py
hfjn: gotcha. thanks for the clarification.

hfjn (approved):

PR #130, opened 2019-12-13
Use offsets_for_times method for setting consumer offset

Bibob7:


Bibob7:
I think there are no tests for the set command, right? Then I would add some additionally.

ognjen-j:
There are some tests, but they are in tests/integration/commands/test_edit.py, because this command was called edit consumergroup. Maybe you can rename the existing ones. But there isn't one for setting the offset to a timestamp.

esque/controller/topic_controller.py
```diff
@@ -114,46 +111,12 @@ def get_cluster_topic(self, topic_name: str) -> Topic:
     def get_local_topic(self, topic_name: str) -> Topic:
         return Topic(topic_name)
 
-    def get_offsets_closest_to_timestamp(self, topic_name: str, timestamp_limit: pendulum) -> Dict[int, int]:
+    def get_offsets_closest_to_timestamp(
```
ognjen-j, line 115: This is basically the same method as offsets_for_timestamp in client.py. Maybe move that one here, and then this one is obsolete?

ognjen-j (approved):

PR #141, opened 2020-02-05
Split folder for key and value schema

### LINKED ISSUE #136: Avro Schema Version is sometimes -1
Some consumed avro messages are stores with schema version -1.
In decode_bytes of AvroMessageDecoder, we are returning -1, if a message has no payload. Maybe we are not handling delete tombstones correctly.

This is fine for consuming and storing, but you want to produce this data, we get an exception, because the schema folder that ends with -1 cannot be find.

Bibob7:
Closes #136

### DIFF
--- a/esque/messages/avromessage.py
+++ b/esque/messages/avromessage.py
@@ -57,7 +57,7 @@ class AvroFileWriter(FileWriter):
         schema_file_path = get_schema_file_path(self.directory, schema_id, schema_folder_name)
         folder_path = schema_file_path.parent
         if not folder_path.exists():
-            folder_path.mkdir()
+            folder_path.mkdir(exist_ok=True)
         schema_file_path.write_text(
             json.dumps(self.schema_registry_client.get_schema_from_id(schema_id).original_schema), encoding="utf-8"
         )
--- a/esque/messages/avromessage.py
+++ b/esque/messages/avromessage.py
@@ -141,7 +141,7 @@ def extract_schema_id(message: bytes) -> int:
 
 def get_schemata(
     base_directory: pathlib.Path, key_schema_id: Optional[int], value_schema_id: Optional[int]
-) -> Tuple[RecordSchema, RecordSchema]:
+) -> Tuple[Optional[RecordSchema], Optional[RecordSchema]]:
     return (
         (None if key_schema_id is None else get_schema(base_directory, key_schema_id, KEY_SCHEMA_FOLDER_NAME)),
         (None if value_schema_id is None else get_schema(base_directory, value_schema_id, VALUE_SCHEMA_FOLDER_NAME)),
--- a/esque/messages/avromessage.py
+++ b/esque/messages/avromessage.py
@@ -8,11 +8,15 @@ from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Tuple
 
 import click
 import fastavro
+from avro.schema import RecordSchema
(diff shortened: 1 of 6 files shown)


esque/messages/avromessage.py
```diff
@@ -122,13 +130,15 @@ def decode_message_from_avro(self, message: Message):
         decoded_message = DecodedAvroMessage(
             decoded_key, decoded_value, message.partition(), key_schema_id, value_schema_id, headers=headers
         )
-        schema_version = it.count(1)
```
Bibob7, line 125: We don't need to consider schema version, because schema ids are already unique identifiers for schemata in schema registry.

swenzel: Also, since there will always be a new counter starting at 1, the schema version doesn't actually change across calls to this method.

esque/messages/message.py
```diff
@@ -127,18 +125,22 @@ def serialize_message(message: Message):
...
-        key=key, value=value, partition=partition, key_schema=key_schema, value_schema=value_schema, headers=headers
-    )
+    return KafkaMessage(key=key, value=value, partition=partition, headers=headers)
+
+
+def get_partitions_in_path(input_directory: pathlib.Path) -> Iterable[str]:
+    path_list = glob(str(input_directory / "partition_*"))
+    partitions = []
+    for partition_path in path_list:
+        partitions.append(partition_path.rpartition("_")[2])
```
swenzel:
```suggestion
        partitions.append(partition_path.rsplit("_", 1)[1])
```

esque/messages/avromessage.py
```diff
@@ -41,29 +47,20 @@ def write_message(self, message: Message):
...
-        directory.mkdir()
-        (directory / "key_schema.avsc").write_text(
-            json.dumps(self.schema_registry_client.get_schema_from_id(key_schema_id).original_schema), encoding="utf-8"
+    def _dump_schema(self, schema_id: int, schema_folder_name: str):
+        if schema_id == -1:
+            return
+        schema_file_path = get_schema_file_path(self.directory, schema_id, schema_folder_name)
+        folder_path = schema_file_path.parent
+        if not folder_path.exists():
+            folder_path.mkdir()
```
MrTrustworthy: `folder_path.mkdir(exists_ok=True)`

esque/messages/avromessage.py
```diff
@@ -122,17 +121,35 @@ def decode_message_from_avro(self, message: Message):
...
         return key_schema_id, value_schema_id, decoded_message, serializable_message
 
 
 def extract_schema_id(message: bytes) -> int:
     magic_byte, schema_id = struct.unpack(">bI", message[:5])
     assert magic_byte == 0, f"Wrong magic byte ({magic_byte}), no AVRO message."
     return schema_id
+
+
+def get_schemata(base_directory: pathlib.Path, key_schema_id: Optional[int], value_schema_id: Optional[int]):
```
MrTrustworthy: Can you annotate the return type?

Bibob7: How can I do this with multiple return values? In this case we would have `RecordSchema, RecordSchema`

hfjn: `Tuple(RecordSchema, RecordSchema)`

swenzel (approved):

swenzel:
Could you also add an entry to the changelog?

PR #148, opened 2020-05-12
Proposal for environment context

hfjn:
This is a proposal for a new environment context which takes precedence of the config file. 
We had the need for a context which can be configured for automation without touching files. 
Happy to hear your feedback.


swenzel:
Hmm instead of going through all that trouble of creating and mapping config variables, you could also create an environment variable `ESQUE_CONFIG_YAML` that can hold the whole esque config as yaml string. That would be more dynamic, future proof and would only require minimal changes.

hfjn:
Hm. I get your point, in general. But that would only get rid of the need to write a file, not to create a YAML in general. :D

swenzel:
You wrote:

> without touching files.

So I thought, that was your problem :smile:

Well, although I don't see how setting 15 environment variables is easier than creating a yaml string, I'm always a fan of "letting the user choose". So I'm not entirely against it.  
I'm just afraid that we might have to rename, add and/or remove some of the variables while our config evolves. We do have a migration mechanism for file based configs but not for environment variables. Do you think a separate command to add a section to the config would help you?  
Something like `esque config add-context foo --bootstrap-servers broker1,broker2 --schema-registry registry ...`  
I think it's easier to keep a command stable than it is to keep the environment variables the same over time. But that's open for discussion :slightly_smiling_face:

hfjn:
Guess I should have been clearer. Sorry for that. :D 
I think you're right with your approach of not wanting to make it unnecessarily complicated and I think the idea of just adding another command is good. I'll see with what I can come up with over the weekend. 😊

PR #151, opened 2020-06-22
Version bump to 0.2.0

hfjn:
Let's go to a proper version number. :) Kept beta state in the `pyproject.toml`. Can change to production/stable if you disagree


swenzel (approved):
removing the beta suffix is fine... I mean we're still sub 1.0.0 release so we don't actually need it.

PR #156, opened 2020-08-27
Disable commits in simple consumers

### LINKED ISSUE #90: Disable Commits in simple consumers
We don't really need consumergroups for the ping consumergroups and probably also for the groups used in #82 and #84.

We can achieve that by [disabling auto-commit in those consumer classes.](https://github.com/confluentinc/confluent-kafka-python/issues/250)

swenzel:
closes #90

### DIFF
--- a/esque/cli/commands.py
+++ b/esque/cli/commands.py
@@ -28,7 +28,7 @@ from esque.cli.output import (
 )
 from esque.clients.consumer import ConsumerFactory, consume_to_file_ordered, consume_to_files
 from esque.clients.producer import PingProducer, ProducerFactory
-from esque.config import PING_GROUP_ID, PING_TOPIC, Config, config_dir, config_path, migration, sample_config_path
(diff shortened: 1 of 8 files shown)


swenzel:
Alright, this is it. I think it (currently) is not possible to bypass the group authorization mechanism with the confluent consumer. Therefore the Kafka user still needs to be authorized to use the corresponding group ids.

We might also want to add a "--resumable" keyword to the "consume" command in the future in order to make use of committed offsets if desired.

Shall we consider renaming the default group id a breaking change? The users need different authorization after all... we could also think about making this configurable via the esque config.

ognjen-j (approved):

garrettthomaskth:
@hfjn you okay if we merge this? :)

hfjn (approved):

hfjn:
Nice one! Merge away.

PR #157, opened 2020-08-27
refactor consume_to_file_ordered

garrettthomaskth:


esque/clients/consumer.py
```diff
@@ -295,13 +298,34 @@ def consume_to_file_ordered(
     output_directory: pathlib.Path,
     topic: str,
     group_id: str,
-    partitions: list,
-    numbers: int,
+    partitions: List[int],
+    desired_count_messages: int,
```
swenzel: `desired_message_count` would sound a little more natural

esque/clients/consumer.py
```diff
@@ -316,63 +340,68 @@ def consume_to_file_ordered(
...
-                    partitions.remove(partition_counter)
-                    if len(partitions) == 0:
-                        messages_left = False
+                # a possible timeout due to a network issue, retry (but not more than MAX_RETRY_COUNT attempts)
+                # retry_count = _handle_empty_message_exception(retry_count)
+                retry_count = retry_count + 1
+                if retry_count > MAX_RETRY_COUNT:
                     keep_polling_current_partition = False
+                    logger.debug(
+                        f"Could not connect to {partition_counter}. "
```
swenzel:
```suggestion
                        f"Could not read from partition {partition_counter}. "
```

esque/cli/commands.py
```diff
@@ -629,7 +629,7 @@ def get_topics(state: State, prefix: str, output_format: str):
     required=False,
 )
 @click.option(
-    "-n", "--numbers", metavar="<n>", help="Number of messages.", type=click.INT, default=sys.maxsize, required=False
+    "-n", "--number", metavar="<n>", help="Number of messages.", type=click.INT, default=sys.maxsize, required=False
```
swenzel, line 632: This is a breaking change. I think we should keep `--numbers` around for compatibility reasons and emit a deprecation warning.

garrettthomaskth: So we agreed to just release a new version with breaking changes, right? CC: @hfjn

hfjn: if you bump the version number accordingly go for it.

swenzel (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### Version 0.2.2
+* Consume option renamed from `numbers` to `number`.
+* Add `edit offsets` command.
+* Bugfix for `OverflowError` in `pretty_duration` function.
```
swenzel, line 4:
```suggestion
* Bugfix for `OverflowError` in `pretty_duration` function.
* use `esque-client` as default group id for all esque commands
* disable group offset commits for ping, consume and describe
```

swenzel (approved):

PR #158, opened 2020-08-27
Edit consumergroup offsets

ognjen-j:


esque/cli/output.py
```diff
@@ -125,11 +125,11 @@ def pretty_duration(value: Any, *, multiplier: int = 1) -> str:
...
     # Fix for conversion errors of ms > C_MAX_INT in some internal lib
-    if value > C_MAX_INT:
-        value = int(value / MILLISECONDS_PER_YEAR)
+    try:
+        value = int(float(value) / MILLISECONDS_PER_YEAR)
         return pendulum.duration(years=value).in_words()
-
-    return pendulum.duration(milliseconds=value).in_words()
+    except OverflowError:
+        return "unlimited"
```
swenzel: If value is not a time range of multiple years, this function should still return `pendulum.duration(milliseconds=value).in_words()`. We might want to add a few unit tests so this is properly covered.

esque/cli/output.py
```diff
@@ -125,11 +125,11 @@ def pretty_duration(value: Any, *, multiplier: int = 1) -> str:
     value *= multiplier
 
     # Fix for conversion errors of ms > C_MAX_INT in some internal lib
-    if value > C_MAX_INT:
-        value = int(value / MILLISECONDS_PER_YEAR)
+    try:
+        value = int(float(value) / MILLISECONDS_PER_YEAR)
```
swenzel:
```suggestion
        value //= MILLISECONDS_PER_YEAR
```

esque/cli/output.py
ognjen-j: I think this should do it.

esque/cli/output.py
ognjen-j: Used the divmod function.

swenzel (approved):
Some ideas for future enhancement: We could support special offset values. Something that allows the user to set the offset to beginning, end or a specific time or to shift it by a specific offset or duration?

PR #160, opened 2020-10-05
Converted the unserializable values to strings.

ognjen-j:


esque/messages/avromessage.py
```diff
@@ -72,7 +73,14 @@ def __init__(self, schema_registry_client):
...
         _, _, _, plaintext_message = self.avro_decoder.decode_message_from_avro(message)
-        click.echo(json.dumps(plaintext_message))
+        click.echo(json.dumps(plaintext_message, default=StdOutAvroWriter.deserializer))
+
+    @staticmethod
+    def deserializer(value):
+        if isinstance(value, (datetime, date)):
+            return value.isoformat()
+        elif isinstance(value, bytes):
+            return "0x{}".format(value.hex())
```
swenzel: How about using base64? That will be a lot shorter for larger messages...  
https://docs.python.org/3/library/base64.html#module-base64

ognjen-j: Sure, why not.

swenzel (approved):

PR #161, opened 2020-10-06
Release prep

ognjen-j:
Version changed to v0.2.3a1


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "esque"
-version = "0.2.2"
+version = "0.2.3a1"
```
swenzel: I think we don't need an alpha version for this. Since we're still on major version zero, just a patch bump is fine.

swenzel (approved):

PR #164, opened 2020-10-21
Expanded commands for deletion to support multiple arguments.

ognjen-j:
The purpose of this PR is to introduce support for multiple command-line arguments for topic and consumer group deletions. The arguments can be provided in three different scenarios:
* only command line arguments are used (`esque delete topic topic1 topic2 topic3`)
* arguments are piped in from a different command (`esque get topics | esque delete topic`)
* both (`esque get topics | esque delete topic topic1 topic2`)

### DIFF
--- a/tests/integration/commands/test_deletion.py
+++ b/tests/integration/commands/test_deletion.py
@@ -33,17 +33,6 @@ def test_topic_deletion_without_verification_does_not_work(
 
 
 @pytest.mark.integration
-def test_delete_topic_without_topic_name_fails(
-    interactive_cli_runner: CliRunner, confluent_admin_client: confluent_kafka.admin.AdminClient
-):
-    n_topics_before = len(confluent_admin_client.list_topics(timeout=5).topics)
-    result = interactive_cli_runner.invoke(delete_topic)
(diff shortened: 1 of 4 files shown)


tests/integration/commands/test_deletion.py
```diff
@@ -32,17 +32,6 @@ def test_topic_deletion_without_verification_does_not_work(
     assert topic in topics
 
 
-@pytest.mark.integration
-def test_delete_topic_without_topic_name_fails(
```
garrettthomaskth, line 36: Why is this test being removed?

ognjen-j: When you pipe arguments from a different command, the current arguments are empty. So the argument is not required anymore.

ognjen-j: I refactored it to handle the result differently.

hfjn:
If you want to test the "piped in" topics via stdin you might be able to mock it with pytest's monkeypatch. :) 
```
def test_method(monkeypatch):
    monkeypatch.setattr('sys.stdin', io.StringIO('topic_1 topic_2'))
```

hfjn:
Otherwise really like it. Feels more unix-like. 👍

swenzel (approved):

PR #176, opened 2021-03-16
Fix consumergroup operations with secured endpoint

### LINKED ISSUE #175: Delete consumergroup only works with PLAINTEXT
The issue is here:
https://github.com/real-digital/esque/blob/a04ccf43e1d840679f74b3f273f88d1d5549419b/esque/controller/consumergroup_controller.py#L53  
We need to pass in more than just the bootstrap servers here. Maybe it's good to have another function like this for the `kafka-python` clients.
https://github.com/real-digital/esque/blob/a04ccf43e1d840679f74b3f273f88d1d5549419b/esque/config/__init__.py#L204

swenzel:
closes #175


Crandel (approved):
LGTM

Crandel (approved):

PR #182, opened 2021-04-15
Add docker container

hfjn:
Added a dockerfile and toolbinding which can act as a drop-in replacement for the locally installed pip version. Should help to nicely decouple esque from the system's python interpreter. 

I don't know how to push to your docker registry. If you could add that we can even provide a ready to use docker image.

### DIFF
--- a/docker_binding.sh
+++ b/docker_binding.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+esque(){
+  docker run \
+    --mount type=bind,source="$(pwd)",target=/home/esque/work \
+    --mount type=bind,source="${HOME}/.esque/config.yaml",target=/home/esque/work/config.yaml \
+    -e ESQUE_CONF_PATH=/home/esque/work/config.yaml \
+    -e PYTHONPATH=/esque \
+    esque "${@}"
+}
--- a/docker_binding.sh
+++ b/docker_binding.sh
@@ -3,8 +3,8 @@
 esque(){
   docker run \
     --mount type=bind,source="$(pwd)",target=/home/esque/work \
-    --mount type=bind,source="${HOME}/.esque/config.yaml",target=/home/esque/config.yaml \
-    -e ESQUE_CONF_PATH=/home/esque/config.yaml \
+    --mount type=bind,source="${HOME}/.esque/esque_config.yaml",target=/home/esque/esque_config.yaml \
+    -e ESQUE_CONF_PATH=/home/esque/esque_config.yaml \
     -e PYTHONPATH=/esque \
     esque "${@}"
 }
--- a/docker_binding.sh
+++ b/docker_binding.sh
@@ -4,7 +4,8 @@ esque(){
   docker run \
     --mount type=bind,source="$(pwd)",target=/home/esque/work \
     --mount type=bind,source="${HOME}/.esque/esque_config.yaml",target=/home/esque/esque_config.yaml \
+    --rm \
+    -it \
     -e ESQUE_CONF_PATH=/home/esque/esque_config.yaml \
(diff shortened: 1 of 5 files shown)


tool.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ && ./configure --prefix /usr \
+ && make \
+ && make install
+
+RUN git clone --branch "1.4.0" --depth 1 https://github.com/edenhill/kafkacat.git \
+ && cd kafkacat \
+ && ./bootstrap.sh \
+ && cp kafkacat /usr/bin/ \
+ && cd .. \
+ && rm -rf kafkacat/
```
swenzel: There are apt packages for both of these. So it should be sufficient to run `apt-get install kafkacat` if you want to include this for debugging.  
The wheels for `confluent_kafka` already contain the necessary libraries, which means for esque itself you don't even need to explicitly install or build librdkafka.

tool.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN mkdir -p /esque
+
```
swenzel:
```suggestion
```

> If the WORKDIR doesn’t exist, it will be created even if it’s not used in any subsequent Dockerfile instruction.

https://docs.docker.com/engine/reference/builder/#workdir

tool.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN poetry config "virtualenvs.create" "false"
+RUN poetry install
+
+# Create user
+RUN useradd -ms /bin/bash esque
+USER esque
+
+ENV PATH="/home/esque/.local/bin:$PATH"
+
+RUN mkdir -p /home/esque/work
```
swenzel:
```suggestion
```
Same as before.

tool.Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM python:3.9
```
swenzel, line 1: I'd also install `vim` so you can run the edit commands.

tool.Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+COPY . /esque
+RUN poetry config "virtualenvs.create" "false"
+RUN poetry install
+
+# Create user
+RUN useradd -ms /bin/bash esque
+USER esque
+
+ENV PATH="/home/esque/.local/bin:$PATH"
```
swenzel: Is this for the `esque` entrypoint or for any other potential tool that is going to be installed?  
Because the `esque` entrypoint is already in the PATH at `/usr/loca/bin` since you install it before you create and switch to the `esque` user which makes it a global install.

docker_binding.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+esque(){
+  docker run \
```
swenzel, line 4: Three suggestions here :slightly_smiling_face: 
1. add `--rm` so the esque containers don't pile up
2. add `-it` so you can communicate with esque
3. remove `PYTHONPATH`, it's not required since `poetry` takes care of that

tool.Dockerfile
hfjn: Fair. I assumed the way in the main `Dockerfile` was the current "working" way.

tool.Dockerfile
hfjn: Hm. Weird I was getting errors which pointed me to the missing path. Can not reproduce that anymore though. So fair enough.

hfjn:
I added a build command for the tool image to the makefile.

tool.Dockerfile
swenzel: yeah, that one is pretty old... we should update it as well.

swenzel (approved):

PR #184, opened 2021-04-15
Don't pull partition data on apply

### LINKED ISSUE #183: Esque apply is incredible slow.
Recently `esque apply` doesn't seem to work properly anymore. It takes several minutes for the initial diff to load.
We are running Kafka 2.4.0. with not too many topics but quite a few partitions.

In Kafka Logs I see a few of these:
```
WARN Attempting to send response via channel for which there is no open connection, connection id
```

I assume that some metadata request is just timing out or taking forever and esque is swallowing that error or actually waiting for something here. What are we waiting for here?

hfjn:
This is an attempt to speed up the apply command.
Problem is that we are retrieving all partitions data every time we touch a topic. Which really impacts the time apply needs to initially load the diff.
I also saw that we scrape each topic twice. Once during initial load and once during the diff. But I left that for now.
This should fix #183.

### DIFF
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -47,7 +47,7 @@ class TopicController:
             topic_names = sorted(topic_names)
 
         if get_topic_objects:
-            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=False) for topic_name in topic_names]
+            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=get_partitions) for topic_name in topic_names]
         else:
             topics = list(map(self.get_local_topic, topic_names))
         return topics
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -47,7 +47,10 @@ class TopicController:
             topic_names = sorted(topic_names)
 
         if get_topic_objects:
-            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=get_partitions) for topic_name in topic_names]
+            topics = [
+                self.get_cluster_topic(topic_name, retrieve_partition_data=get_partitions)
+                for topic_name in topic_names
+            ]
         else:
             topics = list(map(self.get_local_topic, topic_names))
         return topics
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -131,7 +131,9 @@ class TopicController:
     ) -> Topic:
         """Takes a topic and, based on its name, updates all attributes from the cluster"""
 
-        topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp, get_partition_data=retrieve_partition_data)
+        topic.partition_data = self._get_partitions(
+            topic, retrieve_last_timestamp, get_partition_data=retrieve_partition_data
+        )
         topic.config = self.cluster.retrieve_config(ConfigResource.Type.TOPIC, topic.name)
 
         topic.is_only_local = False
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -131,10 +131,7 @@ class TopicController:
     ) -> Topic:
         """Takes a topic and, based on its name, updates all attributes from the cluster"""
 
-        if retrieve_partition_data:
-            topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp)
-        else:
-            topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp, get_partition_data=False)
+        topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp, get_partition_data=retrieve_partition_data)
         topic.config = self.cluster.retrieve_config(ConfigResource.Type.TOPIC, topic.name)
 
         topic.is_only_local = False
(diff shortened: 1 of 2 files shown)


esque/controller/topic_controller.py
```diff
@@ -119,23 +130,34 @@ def get_offsets_closest_to_timestamp(
...
+        self, topic: Topic, retrieve_last_timestamp: bool, get_partition_data: bool = True
+    ) -> List[Partition]:
         config = Config.get_instance().create_confluent_config()
         config.update({"group.id": ESQUE_GROUP_ID, "topic.metadata.refresh.interval.ms": "250"})
         with closing(confluent_kafka.Consumer(config)) as consumer:
             confluent_topic = consumer.list_topics(topic=topic.name).topics[topic.name]
             partitions: List[Partition] = []
+            if not get_partition_data:
+                return [
+                    Partition(partition_id, 0, 0, meta.isrs, meta.leader, meta.replicas, None)
```
swenzel:
```suggestion
                    Partition(partition_id, -1, -1, meta.isrs, meta.leader, meta.replicas, None)
```
I think this will make it more clear, that those aren't actual offsets.

esque/controller/topic_controller.py
```diff
@@ -119,23 +130,34 @@ def get_offsets_closest_to_timestamp(
...
         topic.config = self.cluster.retrieve_config(ConfigResource.Type.TOPIC, topic.name)
 
         topic.is_only_local = False
 
         return topic
 
-    def _get_partition_data(self, topic: Topic, retrieve_last_timestamp: bool) -> List[Partition]:
-
+    def _get_partitions(
+        self, topic: Topic, retrieve_last_timestamp: bool, get_partition_data: bool = True
```
swenzel, line 147: Should there be an assertion to make sure this isn't called with `retrieve_last_timestamp=True` and `get_partition_data=False`?

esque/controller/topic_controller.py
```diff
@@ -119,23 +130,34 @@ def get_offsets_closest_to_timestamp(
...
+    def update_from_cluster(
+        self, topic: Topic, *, retrieve_last_timestamp: bool = False, retrieve_partition_data: bool = True
+    ) -> Topic:
         """Takes a topic and, based on its name, updates all attributes from the cluster"""
 
-        topic.partition_data = self._get_partition_data(topic, retrieve_last_timestamp)
+        if retrieve_partition_data:
+            topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp)
+        else:
+            topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp, get_partition_data=False)
```
swenzel:
```suggestion
        topic.partition_data = self._get_partitions(topic, retrieve_last_timestamp, get_partition_data=retrieve_partition_data)
```

esque/controller/topic_controller.py
```diff
@@ -46,7 +48,10 @@ def list_topics(
             topic_names = sorted(topic_names)
 
         if get_topic_objects:
-            topics = list(map(self.get_cluster_topic, topic_names))
+            if get_partitions:
+                topics = list(map(self.get_cluster_topic, topic_names))
+            else:
+                topics = list(map(partial(self.get_cluster_topic, retrieve_partition_data=False), topic_names))
```
swenzel:
```suggestion
            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=False) for topic_name in topic_names]
```

esque/controller/topic_controller.py
```diff
@@ -2,6 +2,7 @@
 import re
 import time
 from contextlib import closing
+from functools import partial
```
swenzel:
```suggestion
```

esque/controller/topic_controller.py
hfjn: sounds like a good idea. will add that.

hfjn: Done.

esque/controller/topic_controller.py
```diff
@@ -48,10 +47,7 @@ def list_topics(
             topic_names = sorted(topic_names)
 
         if get_topic_objects:
-            if get_partitions:
-                topics = list(map(self.get_cluster_topic, topic_names))
-            else:
-                topics = list(map(partial(self.get_cluster_topic, retrieve_partition_data=False), topic_names))
+            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=False) for topic_name in topic_names]
```
swenzel:
```suggestion
            topics = [self.get_cluster_topic(topic_name, retrieve_partition_data=get_partitions) for topic_name in topic_names]
```
Ups :sweat_smile:

hfjn: :smile:

Crandel:
LGTM

swenzel (approved):

Crandel (approved):

hfjn:
@swenzel @Crandel Can I safely ignore the failed test on 3.8? Looks like a timeout to the test cluster. I guess CI weirdness?

hfjn:
Well I assume that is a yes. :D Thanks!

PR #186, opened 2021-05-27
Refactored the delete topic logic.

ognjen-j:
Added a new command: `delete topics`, while keeping the previous command structure.

### DIFF
--- a/esque/controller/topic_controller.py
+++ b/esque/controller/topic_controller.py
@@ -12,7 +12,8 @@ from confluent_kafka.admin import ConfigResource
 from confluent_kafka.cimpl import KafkaException, NewTopic, TopicPartition
 
 from esque.config import ESQUE_GROUP_ID, Config
-from esque.helpers import ensure_kafka_future_done, ensure_kafka_futures_done
+from esque.errors import TopicDeletionException
+from esque.helpers import ensure_kafka_future_done
 from esque.resources.topic import Partition, Topic, TopicDiff
 
 logger: Logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


Crandel (dismissed):

esque/controller/topic_controller.py
```diff
@@ -96,9 +96,23 @@ def _get_altered_config(self, topic: Topic) -> Dict[str, str]:
...
         return altered_config
 
-    def delete_topic(self, topic: Topic):
-        future = self.cluster.confluent_client.delete_topics([topic.name])[topic.name]
-        ensure_kafka_future_done(future)
+    def delete_topic(self, topic: Topic) -> bool:
+        return self.delete_topics([topic])
+
+    def delete_topics(self, topics: List[Topic]) -> bool:
+        futures = self.cluster.confluent_client.delete_topics([topic.name for topic in topics])
```
swenzel: Looks like there is a way to tell Kafka only to return upon success or timeout. So this might be more efficient because we don't need to keep polling all topics to figure out if they got deleted.

```python
    def delete_topics(self, topics: List[Topic]) -> bool:
        futures = self.cluster.confluent_client.delete_topics([topic.name for topic in topics], operation_timeout=60)
        ensure_kafka_futures_done(list(futures.values()))
        return True
```

Maybe it's also better to iterate over the futures and check their results instead of using `ensure_kafka_futures_done`.
That way we could get more informative error handling. Not all topics might fail and the ones that do, might have some information as to why they failed.

ognjen-j: It does look somewhat faster, subjectively.

swenzel (approved):

PR #188, opened 2021-06-02
Io consumer producer restructuring

Crandel:


esque/cli/commands.py
```diff
@@ -1118,3 +1120,134 @@ def ping(state: State, times: int, wait: int):
...
+    # Extract and format message value using jq.
+    esque io -i "kafka+json:///mytopic" -o "pipe+json://stdout?skip_marker=1" | jq '.value | fromjson'
+
+    \b
+    # Dump messages from a kafka topic into a file
+    esque io -i "kafka+raw://dev/mytopic" -o "pipe+json://stdout" > mytopic_dump.json
+
+    \b
+    # Restore messages from a file
+    cat mytopic_dump.json > esque io -i "pipe+json://stdout" -o "kafka+raw://dev/mytopic"
```
garrettthomaskth: Should this be `"pipe+json://stdin"`?

esque/cli/commands.py
```diff
@@ -1118,3 +1120,134 @@ def ping(state: State, times: int, wait: int):
...
+          The marker line "__esque_msg_marker__\\n" is required for esque to be able to **read** from a pipe.
+          Therefore, you can skip the markers in the output when you're not intending to read the piped data back in.
+          You'll even **need** to skip the marker if you want to further process the messages with tools like `jq`.
+
+    \b
+    Available Serializers:
+    - json
+      Supported Parameters:
+        * indent
+          When serializing data into json, use this many spaces to indent (pretty format) the output.
```
garrettthomaskth, line 1197: By default the serializer uses no line breaks or indentation and puts everything into one line.

esque/cli/commands.py
```diff
@@ -1118,3 +1120,134 @@ def ping(state: State, times: int, wait: int):
...
+    <serializer scheme> - Define data serialization format for key and value in message (e.g json or avro).\b
+                          If no value serializer scheme defined key scheme will be used for both.
+                          (See "Available Serializers" below)
+    <host> and <path>   - Depends on handler and defines the exact source or target of the data pipeline.
+                          (See "Available Handlers" below)
+    <param>             - Define parameters for handler and/or serializer. Use the following prefixes to define
+                          which component a parameter should be used for:
+                            k__  use as key serializer parameter
+                            v__  use as value serializer parameter
+                            kv__ use as key and value serializer parameter
```
garrettthomaskth: It is not obvious to me when a parameter would have the `kv__` prefix

esque/cli/commands.py
```diff
@@ -1118,3 +1120,134 @@ def ping(state: State, times: int, wait: int):
...
+          A subject is the topic name with "-key" or "-value" as suffix, depending on whether this is the key or value
+          serializer.
+
+    \b
+    EXAMPLES:
+    # Read data from an avro topic
+    esque io -i "kafka+reg-avro:///mytopic?kv__schema_registry_uri=$(esque urlencode https://schema-registry.example.com)"
+
+    \b
+    # Extract and format message value using jq.
```
garrettthomaskth: Super small I don't even want to write it 😅 I think if it is a period it should be removed to match the other example descriptions and if it references the `.` you sometimes put after `jq` doesn't there have to be a space in between?

esque/cli/commands.py
swenzel: Is it more clear now? :slightly_smiling_face:

esque/cli/commands.py
swenzel: No it's fine, you're right, now is the time to fix it :smile:

esque/cli/commands.py
swenzel: Of course :man_facepalming:   
Good catch!

esque/cli/commands.py
garrettthomaskth: Yeah! :)

garrettthomaskth (approved):
Looks good to me! :)

PR #190, opened 2021-06-07
Implement esque create consumergroup command

Crandel:


esque/cli/commands.py
```diff
@@ -429,6 +431,30 @@ def edit_offsets(state: State, consumer_id: str, topic_name: str):
...
 
+@create.command("consumergroup")
+@click.argument("consumergroup-id", callback=fallback_to_stdin, required=True, type=click.STRING)
+@click.argument("topic", callback=fallback_to_stdin, required=True, type=click.STRING)
+@click.option("-o", "--offset", required=False, type=click.INT, help="Set offset(default to 0)")
+@click.option("-p", "--partition", required=False, type=click.INT, help="Set partition(default to 0)")
+@default_options
+def create_consumer_group(state: State, consumergroup_id: str, topic: str, offset: int, partition: int):
+    """Create consumer group for topic"""
+    if not ensure_approval("Are you sure?", no_verify=state.no_verify):
```
swenzel: How about adding some more information here? Something like:
```plain
This will create the consumer group <name> with initial offsets:
<topic_name>[<partition>]: <offset>
Are you sure?
```

tests/integration/commands/test_creation.py
```diff
@@ -107,3 +107,17 @@ def test_topic_creation_with_template_works(
     assert config_from_template.num_partitions == num_partitions
     for config_key, value in config.items():
         assert config_from_template.config[config_key] == value
+
+
+def test_consumer_group_creation(
```
swenzel: How about another test that checks the `--partition` and `--offset` parameters?

Thinking about it... maybe it would be good to be able to specify more than one topic partition and offset. How about a multi argument that takes strings in the form:  `<topic name>([<partition>])(=<offset>)` where partition defaults to all and offset defaults to 0? E.g.:
`"mytopic"` -> set offset 0 for all partitions of topic `mytopic`
`"myotpic=10"` -> set offset 10 for all partitions of topic `mytopic`
`"mytopic[0]=10"` -> set offset 10 for partition 0 of topic `mytopic`

esque/cli/commands.py
```diff
@@ -432,25 +433,41 @@ def edit_offsets(state: State, consumer_id: str, topic_name: str):
...
+@click.argument("topics", callback=fallback_to_stdin, required=True, type=click.STRING, nargs=-1)
 @default_options
-def create_consumer_group(state: State, consumergroup_id: str, topic: str, offset: int, partition: int):
-    """Create consumer group for topic"""
-    if not ensure_approval("Are you sure?", no_verify=state.no_verify):
-        click.echo("Aborted!")
+def create_consumer_group(state: State, consumergroup_id: str, topics: str):
+    """Create consumer group for several topics using format <topic_name>[partition]=offset"""
+    offset_pattern = re.compile(r".+=(?P<offset>\d+)$")
+    main_pattern = re.compile(r"^(?P<topic_name>.+)\[(?P<partition>\d+)\].*")
```
swenzel: You can use this pattern to catch all cases :slightly_smiling_face:
https://regex101.com/r/4ydsOB/1
Non-matched groups will just be `None`

Crandel: Thanks!
Done

swenzel (approved):

PR #194, opened 2021-08-03
Add timestamps flag to get latest_timestamps for message

Crandel:


tests/integration/commands/test_describe.py
```diff
@@ -48,7 +48,7 @@ def test_describe_topic_last_timestamp_does_not_commit(
     # cannot use pytest.raises(ConsumerGroupDoesNotExistException) because other tests may have committed offsets
     # for this group
     try:
-        data = consumergroup_controller.get_consumer_group(config.ESQUE_GROUP_ID).describe(verbose=True)
+        data = consumergroup_controller.get_consumer_group(config.ESQUE_GROUP_ID).describe(partitions=True)
```
ognjen-j, line 51: Could you add an assertion for the new field somewhere? Or maybe produce something to two partitions and verify that the timestamps match (or that they are at least different)?

Crandel: Done

esque/resources/consumergroup.py
```diff
@@ -67,55 +68,79 @@ def describe(self, *, verbose=False):
...
+        for partition_id, consumer_offset in partition_data.items():
+            current_offset = consumer_offset
+            old_min, old_max = new_consumer_offsets["consumer_offset"]
+            new_consumer_offsets["consumer_offset"] = (min(old_min, current_offset), max(old_max, current_offset))
+
+            old_min, old_max = new_consumer_offsets["topic_low_watermark"]
+            new_consumer_offsets["topic_low_watermark"] = (
+                min(old_min, topic_watermarks[partition_id].low),
+                max(old_max, topic_watermarks[partition_id].low),
+            )
```
swenzel: How about extracting a method from this? Something like.
```python
N = Union[int, float]  # N for number
def update_minmax(minmax: Tuple[N, N], value: N) -> Tuple[N, N]:
    prev_min, prev_max = minmax
    return (min(prev_min, value), max(prev_max, value))
```
Then you can use it like this:
```python
new_consumer_offsets["topic_low_watermark"] = update_minmax(
    new_consumer_offsets["topic_low_watermark"],
    topic_watermarks[partition_id].low,
)
```

swenzel (approved):

PR #202, opened 2021-09-17
Io create path handler

swenzel:


esque/io/handlers/path.py
```diff
@@ -0,0 +1,106 @@
...
+        self._create_directory_if_necessary()
+        for message in skip_stream_events(message_stream):
+            pickle.dump(message, self._file_output)
+        self._file_output.flush()
+
+    def _create_directory_if_necessary(self):
+        if self._directory_exists:
+            return
+
+        self.config.path_obj.mkdir(parents=True)
```
ognjen-j, line 91: Maybe take into account that the user might not be allowed to write or read in this directory?

swenzel: I've changed the exception handling mechanism so that this case will look like this:

```plain
$ esque io -i "kafka+raw:///mytopic" -o "path+raw:////root/test"
Exception of type PermissionError occurred: [Errno 13] Permission denied: '/root/test'
Run with `--verbose` for complete error.
```

Crandel (approved):

PR #203, opened 2021-09-27
Use io for consume and produce

### LINKED ISSUE #152: Application hangs when trying to read from the end of the stream
When using the --last flag, esque hangs indefinitely. It doesn't read any new messages that come into the topic.

### LINKED ISSUE #155: esque consume ... --preserve-order causes error
For me, running `esque consume <topic-name> --first --stdout` returns messages, while `esque consume <topic-name> --first --stdout --preserve-order` returns `Failed to create socket: Too many open files` and `Failed to resolve '<broker>:9093': nodename nor servname provided, or not known (after 4ms in state CONNECT)`

swenzel:
closes #155
closes #152

### DIFF
--- a/esque/cli/commands/produce.py
+++ b/esque/cli/commands/produce.py
@@ -3,8 +3,10 @@ import pathlib
 import click
 
 from esque.cli.autocomplete import list_contexts, list_topics
+from esque.cli.helpers import ensure_approval
 from esque.cli.options import State, default_options
 from esque.cli.output import blue_bold, green_bold
+from esque.cluster import Cluster
 from esque.io.handlers import BaseHandler, KafkaHandler, PathHandler, PipeHandler
 from esque.io.handlers.kafka import KafkaHandlerConfig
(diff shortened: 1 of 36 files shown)


Crandel (dismissed):

Crandel (dismissed):

esque/cli/commands/produce.py
```diff
@@ -96,55 +108,60 @@ def produce(
...
+
+    output_handler = create_output_handler(to_context, topic)
+    builder.with_output_handler(output_handler)
+
+    total_number_of_messages_produced = 0
+
+    if match:
+        builder.with_stream_decorator(yield_only_matching_messages(match))
+
+    def counter_decorator(message_stream: MessageStream) -> MessageStream:
```
ognjen-j: This looks like the same decorator from the consume command? Since we'll probably need it for transfer as well, how about extracting it?

swenzel: Yeah, good point :+1:

ognjen-j (approved):

PR #227, opened 2025-02-11
Add protobuf support

poursadeqi:
This is the first alpha version of supproting protobuf. Here is the list of changes: 

* packages are updated.
* minimum supported python version is now 3.9 and it's going to support until 3.12.
* pre-commit-config is updated
* docker files and docker-compose files  are updated.
* struct and protobuf serializers are added.
* pyproject.toml is updated with latest version of packages.
* tox is updated to test for py39,py310,py311,py312
* stdout parameter is removed and by default output is going to be stdout
* readme is update.


esque/cli/commands/consume.py
```diff
@@ -58,19 +81,70 @@
...
-    default=False,
-    is_flag=True,
+    "-k",
+    "--key-serializer",
+    type=click.Choice(["s", "str", "b", "binary", "a", "avro", "p", "proto", "struct"], case_sensitive=False),
+    help="Specify deserialization for keys. if you choose avro or binary value will also be set the same unless you choose differently.",
+    default=None,
+)
+@click.option(
+    "-s",
```
i-emek, line 97: An optional note, but nothing big. It might be a little bit confusing to have -k for for the key serializer, and -s for the value serializer. Probably something like -ks and -vs would be less confusing. But the confusion will be resolved once the user get used to the command. Feel free to skip the note

i-emek: The changes look good. I couldn't spot anything concerning. Good job Ebrahim on an adding the support for proto consumption. The feature is really helpful.

esque/io/serializers/proto.py
```diff
@@ -0,0 +1,58 @@
...
+    protoc_py_path: str
+    module_name: str
+    class_name: str
+
+    def __post_init__(self):
+        self._message_class = self._load_message_class()
+
+    def _load_message_class(self) -> Type[Message]:
+        sys.path.append(self.protoc_py_path)
+        module = importlib.import_module(self.module_name)
```
i-emek: One last question. Will ever be there a conflict in case a module is having the same name a library that esque is using?

poursadeqi: I wil rename it to grpc_module just in case

esque/cli/commands/consume.py
poursadeqi: I can rename them to -sk and -vs, your call! But my main focus was making the command easy and quick to use. If I run this command daily, I’d rather type -s or -k and memorize them instead of writing ks or vs—just being lazy! 😆

Also, one more thing: for something like Avro, when I use -s a, the serialization applies to both the key and value unless the key is explicitly set to something else.

i-emek: All good then, we can keep them the way they are now

i-emek (approved):
PR approved

